PR #2, opened 2015-03-18
[admin] Fixing /report

PhrozenByte:
The ```/report``` GUI doesn't disable keybinds, so (for example) typing a ```t``` into one of the input fields fires the chat bind and steals input focus. If you close the ```/report``` GUI while the admin panel is visible in the background, the mouse disappears and leaves a unusable admin panel. After successfully submitting a message, a small window (```aMessageBox```) is shown. The window also shows an Ok-button, which isn't reachable, because the mouse is already hidden. This misleads users to press ```Enter```, what is often a bound key (e.g. ```/kill``` in the race gamemod), even though the window is automatically hidden after 3 seconds.

**Steps To Reproduce**
1. load & start admin resource
2. login as admin
3. open admin panel (default: ```p```)
4. type ```/report```
5. focus is stolen when typing a bound key (e.g. ```t```) into one of the input fields (addressed by the first commit)
6.1. when closing the ```/report``` window, the mouse disappears while the admin panel is still visible (addressed by the second commit)
6.2. when submitting, a success message is shown as described above (addressed by the third commit)

**Mantis**
First commit: http://bugs.mtasa.com/view.php?id=8791
Second commit: -
Third commit: -

### DIFF
--- a/[admin]/admin/client/gui/admin_report.lua
+++ b/[admin]/admin/client/gui/admin_report.lua
@@ -38,6 +38,10 @@ function aReport ( player )
 	end
 	guiBringToFront ( aReportForm )
 	showCursor ( true )
+	
+	if ( aAdminForm == null or guiGetVisible ( aAdminForm ) == false ) then
+		guiSetInputMode ( "no_binds_when_editing" )
+	end
 end
 addCommandHandler ( "report", aReport )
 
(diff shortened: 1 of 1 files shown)


PhrozenByte:
More than a year later neither merged nor rejected or at least commented... :disappointed:

PR #6, opened 2015-04-23
[freeroam] Warp to player with hex code doesn't work

TAPL-93:
Warp to players with hex codes in their name doesn't work when "removeHex" is set to true in the meta.


sbx320:
I think this may cause errors when someone disconnects while you have the teleport window open.

An additional check for 

    isElement(leaf.player)

before warping should fix that.

sbx320:
Looks good to me now.

PR #9, opened 2015-04-25
[freeroam] Fixed camera stuck and gravity mess up when teleport

TAPL-93:


TAPL-93:
The whole thing here is about teleporting more than once before the timers finished, the first time getGravity will return the current gravity and the gravity will be set to 0.001 but the second time getGravity won't return your original gravity but rather will return 0.001, that's why i made grav global variable and used 'if not grav' to make sure it not been set before, therefore setting grav to nil in the timer function is necessary.

ccw808:
Ah yes, I can see it now. Thanks.

PR #13, opened 2015-07-10
Updated joinquit and playerblips

nokizorque:
[joinquit]
- Added a colour control variable for all three RGB chatbox outputs.

[playerblips]
- Added team functionality.
- Rewrote to make use of MTA's functions.
- Simplified and removed many functions.


nokizorque:
I screwed up my commits badly. Need to refork and reorganize.

PR #17, opened 2015-07-20
Code Cleaning

asm95:
Someone copied Control Interface from editor to use in move_keyboard. Though, most features included in this "key binding enhancement package" are not necessary and just made editor slower.

An example is each time the editor suspend/resume, the script does a deep copy of keybinds table and unbind/bind each key of the original one.

Those changes aimed to make the code clear, smaller and somewhat faster.


ArranTuna:
The benefits of code cleaning are negligible, the risk of this change introducing a bug far outweighs the benefits.

PR #20, opened 2015-08-16
Fixed instant reload on sawnoff

Dutchman101:
This glitch was performed by switching to same weapon again, (/wp sawnoff) while in the very beginning of reload anim, cancelling it and evading it. If done quickly and fluently, switching between that /wp bind and constantly firing, will make you able to fire sawn-off rounds constantly and like an automatic, something disastrous seen the damage sawnoff shots do. That is now fixed and prevented with this, adding a forced reload on swapping sawnoff to interrupt the glitcher from progressing to constant firing & repeatedly cancelling reload.

### DIFF
--- a/[gameplay]/freeroam/fr_server.lua
+++ b/[gameplay]/freeroam/fr_server.lua
@@ -283,6 +283,7 @@ function warpMe(targetPlayer)
 	setCameraInterior(source, interior)
 end
 
+local sawnoffAntiAbuse = {}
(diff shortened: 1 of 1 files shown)


qaisjp:
PR continued in https://github.com/multitheftauto/mtasa-resources/pull/36

qaisjp:
PR continued in https://github.com/multitheftauto/mtasa-resources/pull/37

PR #23, opened 2015-08-20
[playerblips] Rewrote and added functionality

nokizorque:


JR10:
A 500ms infinite timer? Nope.

PR #24, opened 2015-08-25
Fix for rapid shooting glitch with slowfiring weapons

Dutchman101:
Because the weapon's defined shooting speeds weren't applied if the method to change a weapon during driveby was to directly set the weapon into wield slot, it ends up as an abusable glitch where you can get for example, Deagle to fire 7 rounds in around a second, this is disastrous seen the damage it does per shot.


Dutchman101:
Thanks. done.

PR #25, opened 2015-09-09
Add a link to the bugtracker

JR10:
A very trivial addition really, might be useful for whoever looks for the "Issues" tab and fails to find it.


qaisjp:
:+1:

PR #26, opened 2015-09-10
[mapmanager] Fix for otherMapsContainer not having a header label

JR10:
To reproduce this make sure that all maps have a gamemode attribute in the info tag
in meta.xml (no maps with no gamemode set) then just click on the mapmanager in the
resourcebrowser

The error was: "Uncaught TypeError: Cannot read property 'style' of undefined"


qaisjp:
What line was the error?

JR10:
[Line 158](https://github.com/JR10/mtasa-resources/blob/pr/mapmanager-fix/%5Bmanagers%5D/mapmanager/mapmanager.js#L158)

qaisjp:
Good job.

PR #31, opened 2015-11-20
[freeroam] Fix for debug warning when selecting a weapon category.

TAPL-93:


qaisjp:
Nice one, we could also disable the accept control when categories are selected.

PR #32, opened 2015-11-22
[playerblips] Added team functionality

nokizorque:
- Added team functionality, with an optional switch.
- Settings for blip size and alpha.

### DIFF
--- a/[gameplay]/playerblips/meta.xml
+++ b/[gameplay]/playerblips/meta.xml
@@ -1,3 +1,4 @@
 <meta>
   <script src="playerblips.lua" type="server" />
+  <oop>true</oop>
 </meta>
--- a/[gameplay]/playerblips/playerblips.lua
+++ b/[gameplay]/playerblips/playerblips.lua
@@ -1,71 +1,76 @@
-﻿-- needs configurable blip colors, and team support
-root = getRootElement ()
-color = { 0, 255, 0 }
-players = {}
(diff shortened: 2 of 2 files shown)


qaisjp:
Thank you for your contribution! The code looks fine from here, but I'll get this tested by the end of the week.

qaisjp:
I've tested and looked through your update and there are a few things that need improving/fixing:
* an error message when a player is not in a team
* team colour needs to default to off (to behave the same out of the box)
* disabling both commands when in team mode (custom colours and team colours mixed sound a bit problematic design wise, IMO)

I think supporting all the different cases would make the resource unnecessarily complex and I don't think that this resource really needs these features and settings right now.

If you can address these issues and explain why you feel that these features should be available then feel free to reopen this pull request, but for now I am closing this pull request.

Thanks again :)

nokizorque:
> - an error message when a player is not in a team

Yeah, that was a poor move on my part.

> - team colour needs to default to off (to behave the same out of the box)
- disabling both commands when in team mode

Fixed!

In the first line of the current playerblips.lua, it says:
> needs configurable blip colors, and team support

I still wanted to preserve the original functionality and add this new functionality. I didn't think of the mixed types though. I guess I'll need to do that better next time.

qaisjp:
Perhaps you could allow those features/settings to be configured using the settings system?

This allows administrators to change the team mode using the admin panel.

PR #35, opened 2015-12-12
Added weapons table to replace old method

alberto2345:
- Replaced the old method of retrieving weapon names with a table to get
rid of bad weapon names such as Bomb and Freefalling Bomb and duplicated Dildo and Vibrator names. (temp fix until getWeaponNameFromID is fixed)

### DIFF
--- a/[admin]/admin/client/gui/admin_main.lua
+++ b/[admin]/admin/client/gui/admin_main.lua
@@ -19,6 +19,50 @@ aBans = {}
 aLastSync = 0
 aResources = {}
 
+local weaponsTable = {
+	"Fist",
+	"Brassknuckle",
+	"Golfclub",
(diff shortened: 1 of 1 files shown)


qaisjp:
> temp fix until getWeaponNameFromID is fixed

Please can you link the bug report for that function?

alberto2345:
https://bugs.mtasa.com/view.php?id=7968

It doesn't say anything about the IDs that say Bomb and Freefalling bomb though.

qaisjp:
Please add a note to the report describing those other problems.

alberto2345:
Ok no problem, I added a note to the issue.

qaisjp:
A better solution would be to blacklist projectile IDs.

PR #37, opened 2015-12-14
Fixed instant reload on sawnoff

Dutchman101:
This is a second remake of old PR: #20 due to repo mismatch. (other remake failed due to some unintended commit linking)

Glitch was performed by switching to same weapon again, (/wp 26) (sawnoff) while in the very beginning of reload anim, cancelling it and evading it. If done quickly and fluently, switching between that /wp bind and constantly firing, will make you able to fire sawn-off rounds constantly and like an automatic, something disastrous seen the damage sawnoff shots do. That is now fixed and prevented with this, adding a forced reload on swapping sawnoff to interrupt the glitcher from progressing to constant firing & repeatedly cancelling reload.

New PR includes the quithandler ccw suggested for it to fit coding standards, and avoiding isElement (we have a proper quithandler now) so that second change included in old PR wasnt added now.

### DIFF
--- a/[gameplay]/freeroam/fr_server.lua
+++ b/[gameplay]/freeroam/fr_server.lua
@@ -286,6 +286,7 @@ function warpMe(targetPlayer)
 	end
 end
 
+local sawnoffAntiAbuse = {}
 function giveMeWeapon(weapon, amount)
 	if weapon and weapon > 50 then
 		return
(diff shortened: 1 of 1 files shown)


qaisjp:
Wait a second, shouldn't this be a part of [setGlitchEnabled("quickreload", ..)](https://wiki.multitheftauto.com/wiki/SetGlitchEnabled)

Dutchman101:
the sawnoff fix is unrelated from setGlitchEnabled because that was performed by simpler switching weapons (GTA/MTA bug) and this one relies on /wp command only found in freeroam that facilitates the glitch to happen, thus is a freeroam glitch.

PR #38, opened 2016-01-11
[playerblips] Added team functionality

nokizorque:


[gameplay]/playerblips/playerblips.lua
```diff
@@ -1,71 +1,73 @@
...
-function onPlayerWasted ( totalammo, killer, killerweapon )
-	destroyBlipsAttachedTo ( source )
+function setBlipColor(plr, _, r, g, b)
+	if (useTeams) then return end
+	if (tonumber(b)) then
+		colors[plr] = {r, g, b}
+		createPlayerBlip(plr)
+	end
 end
+addCommandHandler("setblipcolor", setBlipColor)
```
Comment (ArranTuna) [inline [gameplay]/playerblips/playerblips.lua]: Wouldn't need if (useTeams) then return end if addCommandHandler was added inside the: if (useTeams == "true") then statement. Same for the other command handler.

[gameplay]/playerblips/playerblips.lua
```diff
@@ -1,71 +1,73 @@
...
-			else
-				createBlipAttachedTo ( player, 0, 2, color[1], color[2], color[3] )
-			end
+function setBlipsColor(_, _, r, g, b)
+	if (useTeams) then return end
+	if (tonumber(b)) then
+		color = {tonumber(r), tonumber(g), tonumber(b)}
+		for i, plr in ipairs(Element.getAllByType("player")) do
+			createPlayerBlip(plr)
 		end
```
Comment (ArranTuna) [inline [gameplay]/playerblips/playerblips.lua]: Could do with some error message saying couldn't change blips because invalid arguments.

nokizorque:
Thanks @ArranTuna. I added your suggested improvements.

ArranTuna:
'/setblipscolor' is weird. It allows any player, to change the blip color of all players, which makes no sense. It should either be for admins only or removed.

qaisjp:
I agree with @ArranTuna.  I can't see why blip colors need to be changed on the fly -- the setting should  be sufficient. I think the command should be removed.

ArranTuna:
setBlipColor fails to check if specified arguments are within 0 and 255.

PR #41, opened 2016-02-28
Patch for a better set Player skin

Einheit-101:
This is an easier way of doing the job and it doesnt trigger the "on[Client]PlayerSpawn" events.


Einheit-101:
You are right. I've corrected this.

ccw808:
Thanks, although `mdata = data` is still missing

Einheit-101:
Phew, so many small things :D
That one is better.

qaisjp:
@Einheit-101 - please don't use the default commit title. Keep commit messages constructive and helpful. In the log these updates won't give any kind of hint about what changes were made.

Einheit-101:
I guess like that.

ccw808:
Great, thanks.

PR #42, opened 2016-03-29
Added ability to change the color of each player Blip

Dante383:
As above.


qaisjp:
I don't think this feature is necessary right now. Thank you for your contribution.

Dante383:
I'm added this because comment in 1 line, but as you like.

PR #43, opened 2016-05-05
[missiontimer] fix for "server triggered client side event XYZ but event is not added clientside"

Einheit-101:


missiontimer_server.lua
```diff
@@ -0,0 +1,147 @@
+﻿local thisResource = getThisResource()
+local bool = { [false]=true, [true]=true }
```
Necktrox: You can use `type(var) == "boolean"` instead

missiontimer_server.lua
```diff
@@ -0,0 +1,147 @@
+﻿local thisResource = getThisResource()
```
Necktrox: MTA offers a predefined variable called `resourceRoot`, which you could've used

missiontimer_server.lua
```diff
@@ -0,0 +1,147 @@
...
+
+addEventHandler("onPlayerQuit",root,
+	function()
+		table.removevalue( readyPlayerList, source )
+	end
+)
+
+addEventHandler ("onResourceStop",root,
+	function()
+		for i,timer in ipairs(getElementsByType("missiontimer",source)) do
```
Necktrox: Replace `ipairs` with `pairs`, because it's faster and the order does not matter

Einheit-101:
Yup, all that stuff doesnt belong to me, its from the original author of missiontimer. I implemented your modifications.

qaisjp:
Thanks for the PR. the content hasn't been checked because you have not edited the existing resource, you have to uploaded the file to the root of your repository.

See: https://github.com/Einheit-101/mtasa-resources/tree/(missiontimer)

Please submit a new pull request with this fix. Thanks!

PR #47, opened 2016-06-14
Fix anonymous-admin option to really be anonymous

Dutchman101:
This fixes several occasions on which the admin's nickname would be exposed and shown to the victim.
After this commit, the Anonymous admin tickbox really hides all occasions on which their name would not be hidden, as when the bug existed.


qaisjp:
I don't think anonymous admin should be truly anonymous (in regards to banPlayer) as this could be abused... perhaps you could use setBanAdmin (after `banPlayer`) to preserve the log of who actually banned who? That way the person being banned won't see who banned them either.

What do you think?

PR #48, opened 2016-07-25
Add GUI options on serial / IP ban: Reason & nick

Dutchman101:
Implement possibility to manually add ban reason & player nick when banning a specific serial / IP of offline player through adminpanel ''Ban Serial'' or ''Ban IP'' button.
Previously, customly entering bans per serial / IP in adminpanel resulted in reason-less / nick-less bans that only contained the serial or IP in banlist.

This PR adds 2 fields to the ''Ban serial'' and ''Ban IP'' buttons to enter reason & nickname.


qaisjp:
Nice work!

PR #59, opened 2016-11-22
Fixed /anim abusable glitch

Dutchman101:
This patch blocks 2 anims from being used directly by /anim command rather than GUI, because they were abusable in combat with other players; the anim makes their ped go underground (under the world) so that they suddenly become invisible and unharmable by gunfire since within split second they'd be gone from sight (in reality just under the attacker's feet)

It could be used by command at the moment player almost dies with low HP, to prevent any attack from other player to succeed. It's unfair advantage.

### DIFF
--- a/[gameplay]/freeroam/fr_client.lua
+++ b/[gameplay]/freeroam/fr_client.lua
@@ -114,6 +114,11 @@ wndAnim = {
 
 addCommandHandler('anim',
 	function(command, lib, name)
+		if lib and name then
+		if string.lower(lib) == "finale" and string.lower(name) == "fin_jump_on" or string.lower(lib) == "finale2" and string.lower(name) == "fin_cop1_climbout" then
+			return outputChatBox ("* This animation may not be set by a command", 255, 0, 0)
+			end
+		end
 		server.setPedAnimation(g_Me, lib, name, true, true)
 	end
 )
--- a/[gameplay]/freeroam/fr_client.lua
+++ b/[gameplay]/freeroam/fr_client.lua
@@ -114,10 +114,12 @@ wndAnim = {
 
 addCommandHandler('anim',
 	function(command, lib, name)
-		if lib and name then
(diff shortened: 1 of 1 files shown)


Dutchman101:
To reproduce the glitch;
1) /bind 1 anim FINALE2 FIN_Cop1_ClimbOut and /bind 2 anim FINALE FIN_Jump_On
2) Press bind-key 1 and/or 2 whenever it's too hot under your feet while in a fight

Try again with the patch and they won't work by command, only the anim GUI, which is too time costly to select in a fraction of a second while fighting for it to still be abusable through GUI setting.

qaisjp:
I really think something like this should be set by the server and not a default.

Dutchman101:
qaisjp I don't agree, can you explain why?
Not a single freeroam server would benefit from glitch abuse, these are the only 2 anims out of countless that can be abused in this way.

The only purpose for these anims by command are glitch abusing, it's booming on freeroam servers atm.

qaisjp:
Okay. Could this be implemented in a similar style to how vehicles/weapons are prohibited?

Dutchman101:
Since there's only 2 anims out of hundreds that exist abusable in this way, I think it won't ever happen a player needs to use these 2 specific animations by command for other reasons than abuse, that's just highly unlikely.
It could be like that but then as sort of opt-out, aka the anims are already listed to glitch-fix by default. But seen legitimate neccesary use of these anims by command is very unlikely that would be very unnecesary in my opinion.

qaisjp:
I've made a tweak to the code to make it behave more similar to how other errors do, and to fix some indentation.

I still think that this should be added as a setting on an anim by anim basis, similar to how weapons/vehicles are prohibited. Other animations may want to be disabled for other reasons, namely the `paulnmac` group, for whatever reason. The setting would make it easier for server owners to disable animations (and those two animations could be a default).

I'll leave it to the other developers to make a decision if you have no further arguments for why these animations should be hardcoded into freeroam.

PR #60, opened 2016-12-28
Added freeroam anti-command spam to block crafted attack methods to cause lag

Dutchman101:
_**Note from qaisjp: please squash merge if you decide to squash merge this**_

Players can use a method to load server and cause lag by using several binds **(or macros with many binds attached at once)** to mass-trigger functions that are serverside and many expensive, this was fixed by implementing anti-command spam that cancels execution in an early stage when triggered. ~~The basic anti command spam only limits certain commands to once in 2 secs~~ (and some funcs not, like repair, flip etc to not interfere with gameplay experience that much, as they can be spammed by players for legit reasons) +

- extended detection of deliberate spam, which only triggers at inhumane trigger speeds or mashing a bind, and then applies a command use ''ban'' preventing them to use any of the commands protected by timer, for 10s

More on the attack methods: either several players using binds (or macros) in sequence with multiple server funcs in teamwork at the same moment will lag the server with as source the freeroam resource causing extensive load, they can also use macros on many binds while alone to already cause significant lag spikes. Multiple players using macro binds is even more desastrous and could time everyone out or crash the server potentially.

Server owners: if your server had lag problems and performancebrowser/IPB shows freeroam peaking in load at those moments, chances are you were hit with this misuse this **PR has now fixed.**

### DIFF
--- a/[gameplay]/freeroam/fr_client.lua
+++ b/[gameplay]/freeroam/fr_client.lua
@@ -11,42 +11,58 @@ guiSetInputMode("no_binds_when_editing")
 -- Place to store the ticks for anti spam:
 local antiCommandSpam = {}
 
--- Additional protection:
--- Settings:
-local command_ddos_protection = true
-local tries_required_to_trigger = 15
-local duration_of_global_ban = 10000 -- in milliseconds
+-- Settings are stored in meta.xml
+addEvent("spamProtectionSettings", true)
+function spamProtectionSettings(settings)
+	if settings then
+		command_ddos_protection = settings.command_spam_protection
+		tries_required_to_trigger = settings.tries_required_to_trigger
+		tries_required_to_trigger_exceptions = settings.tries_required_to_trigger_low
+		duration_of_global_ban = settings.command_spam_ban_duration
+	end
+end
+addEventHandler("spamProtectionSettings", localPlayer, spamProtectionSettings)
 
 -- Store the tries for forced global cooldown
-local command_tries = 0
 local global_cooldown = 0
-function isCommandOnCD(cmd)
+function isCommandOnCD(cmd, exception)
 	local tick = getTickCount()
 
 	-- check if a global cd is active
-	if command_ddos_protection and global_cooldown ~= 0 then
+	if command_ddos_protection == "true" and global_cooldown ~= 0 then
 		if tick - global_cooldown <= duration_of_global_ban then
-			outputChatBox ("* You are banned from using commands for "..math.ceil((duration_of_global_ban/1000)-((tick-global_cooldown)/1000)).." more seconds due to continuous spam*", 255, 0, 0)
+			errMsg ("* You are banned from using commands for "..math.ceil((duration_of_global_ban/1000)-((tick-global_cooldown)/1000)).." more seconds due to continuous spam*")
 			return true
 		end
 	end
 
-	if antiCommandSpam[cmd] and tick-antiCommandSpam[cmd] < 2000 then
-		if command_ddos_protection then
-			command_tries = command_tries + 1
-			if command_tries >= tries_required_to_trigger then
-				-- activate a global command cooldown
-				global_cooldown = tick
-				command_tries = 0
-			end
+	if command_ddos_protection == "true" then
+		if not antiCommandSpam[cmd] then
+			antiCommandSpam[cmd] = {time = tick, tries = 1}
+			return false
 		end
-		outputChatBox ("* Failed, don't spam the '"..tostring(cmd).."' command! *", 255, 0, 0)
+
+		local oldTime = antiCommandSpam[cmd].time
+
+		if tick-oldTime > 2000 then
+			antiCommandSpam[cmd].time = tick
+			antiCommandSpam[cmd].tries = 1 
+			return false
+		end
+
+		antiCommandSpam[cmd].tries = antiCommandSpam[cmd].tries + 1
+
+
+		if exception and antiCommandSpam[cmd].tries < tries_required_to_trigger_exceptions then return false end
+
+		if exception == nil and antiCommandSpam[cmd].tries < tries_required_to_trigger then return false end
+
+		-- activate a global command cooldown
+		global_cooldown = tick
+		antiCommandSpam[cmd].tries = 0
+		errMsg ("* Failed, don't spam the '"..tostring(cmd).."' command! *")
 		return true
 	else
-		if command_ddos_protection then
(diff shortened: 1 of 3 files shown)


Dutchman101:
The design of timer is like that to leave room for other integrations I was planning to propose later, as multiple other commands/funcs are serverside and can be abused by the same method, although less expensive/lag effect

Dutchman101:
@qaisjp 
''however, some changes needs to be made''

Like?

ArranTuna:
Fails to null timer. Also the timer spam is inefficient, you can use getTickCount() and store the last time they used a command.

qaisjp:
@Dutchman101 

> Like?

Please refer to the review.

Dutchman101:
Added commit based on Arran's suggestions (code review I cant view any specific comments qaisjp as none listed) how about the new method?

qaisjp:
Sorry about that, here is the (partially outdated) review. Turns out that can work out on a review and submit it when you are finished reviewing.

[gameplay]/freeroam/fr_client.lua
```diff
@@ -8,15 +8,20 @@ g_Me = getLocalPlayer()
...
+-- Place to store the ticks for anti spam:
+local antiCommandSpam = {}
+
+function isCommandOnCD(cmd)
+	local tick = getTickCount()
+	if antiCommandSpam[cmd] and tick-antiCommandSpam[cmd] < 2000 then
+		outputChatBox ("* Failed, don't spam the '"..tostring(cmd).."' command! *", 255, 0, 0)
+		return true
+	else
+		antiCommandSpam[cmd] = tick
```
Comment (qaisjp) [inline [gameplay]/freeroam/fr_client.lua]: I feel like not resetting the timer on failure would be less irritating to users. What do you think?

Comment (Dutchman101) [inline [gameplay]/freeroam/fr_client.lua]: It aint, because mashing a command bind makes you just go through once in 2 seconds.
They dont got to ''start over'' waiting on failure, if thats what you mean.
Try it, smash cmd bind for 10 seconds straight, your command gets passed 5 times in 10 secs, no matter how rapid you hit it.

Comment (Dutchman101) [inline [gameplay]/freeroam/fr_client.lua]: @qaisjp your comment had me thinking and eventually come up with even a better mechanism to separate legitimate players just wanting to speed a process up from deliberate spammers, 1) as said before legitimate ones don't actually deal with a setback in timer reset, however ontop of that I just added a improvement commit to detect deliberate spammers and actually impose such limitations by ignoring their requests for 10 seconds when confirmed malicious (check the video in commit description; https://www.youtube.com/watch?v=ZuCyICDeOhg to see only inhumane triggering speeds or macro's would achieve that))

Dutchman101:
As said the design of timer was made to fit multiple protected commands and I just added several capable of inducing server lag through the same attack method, to the protection.
They got the same 2 seconds spam timer applied and deliberate/macro command trigger detection.

This was put in a separate commit to leave open room for discussion on these additions. I don't think these specific commands added will see any significant gameplay impact on limiting them to once per 2 seconds.

qaisjp:
At this point you can safely just rename the issue to "Implement freeroam spam protection" -- am I correct in thinking this is the general idea of your pull request (at this stage)?

[gameplay]/freeroam/fr_client.lua
```diff
@@ -8,6 +8,49 @@ g_Me = getLocalPlayer()
...
+	if antiCommandSpam[cmd] and tick-antiCommandSpam[cmd] < 2000 then
+		if command_ddos_protection then
+			command_tries = command_tries + 1
+			if command_tries >= tries_required_to_trigger then
+				-- activate a global command cooldown
+				global_cooldown = tick
+				command_tries = 0
+			end
+		end
+		outputChatBox ("* Failed, don't spam the '"..tostring(cmd).."' command! *", 255, 0, 0)
```
Comment (qaisjp) [inline [gameplay]/freeroam/fr_client.lua]: Please use the `errMsg` function.

[gameplay]/freeroam/fr_client.lua
```diff
@@ -8,6 +8,49 @@ g_Me = getLocalPlayer()
...
+-- Store the tries for forced global cooldown
+local command_tries = 0
+local global_cooldown = 0
+function isCommandOnCD(cmd)
+	local tick = getTickCount()
+
+	-- check if a global cd is active
+	if command_ddos_protection and global_cooldown ~= 0 then
+		if tick - global_cooldown <= duration_of_global_ban then
+			outputChatBox ("* You are banned from using commands for "..math.ceil((duration_of_global_ban/1000)-((tick-global_cooldown)/1000)).." more seconds due to continuous spam*", 255, 0, 0)
```
Comment (qaisjp) [inline [gameplay]/freeroam/fr_client.lua]: Please use the `errMsg` function.

[gameplay]/freeroam/fr_client.lua
```diff
@@ -8,6 +8,49 @@ g_Me = getLocalPlayer()
 server = createServerCallInterface()
 guiSetInputMode("no_binds_when_editing")
 
+-- Place to store the ticks for anti spam:
+local antiCommandSpam = {}
+
+-- Additional protection:
+-- Settings:
```
Comment (qaisjp) [inline [gameplay]/freeroam/fr_client.lua]: These should be proper `meta.xml` settings. Evaluate whether setting values should be sent to the client (retrieved server side using `get`) or whether command blocks should be implemented serverside.

Dutchman101:
I've rewritten PR title & description as the current implemented features exceeded the previous, as qaisjp said, and once I get to i'll add the remaining fixes to fit the coding guidelines as outlined by @qaisjp in last review.

Dutchman101:
Just added commits to follow recommendations by @qaisjp and improve logic, it will now even impact gameplay less than before (which was already minimal) by not imposing a command use limit of once each 2 seconds, but instead only triggering anti-command spam if massive spam has already occured (a defined amount of times within 2 seconds timeframe) only then it will suspend/ignore any calls from the client, resulting in migitation of the lag effect on the victim server.

Any thoughts on this new approach too?

qaisjp:
Use [this link](https://github.com/multitheftauto/mtasa-resources/pull/60/commits/e858bb39e66d7b3b09872a2dc38be62f4feaa7c4?w=1) to see a more readable version of my most commit.

Dutchman101:
Thanks, but @qaisjp that causes some bugs like
You are banned from using commands for -2285 seconds due to continuous spam

and contains typo  if command_ddos_protection != "true" then
.. ~=
I'm not near my pc with git/editor right now so you could fix that commit?

qaisjp:
Whoops. Fixed that now. Also should have fixed the maths.

PR #61, opened 2017-01-03
Admin reporting system improvements

Dezash:
Added report category settings to reports.xml configuration. This will allow server owners to configure the report categories.

Added player report functionality. Each report category in reports.xml has an attribute called "isPlayerReport". If set to true, when reporting, the player will be asked to select the player that he wants to report. Admins will later be able to easily review the suspected player's information such as IP address, serial, MTA version, name and chat log.

Added Ban serial and Ban IP buttons in messages viewer. These buttons will only be enabled for report categories that have isPlayerReport set to true.
When clicked, a ban input box will appear with serial or IP automatically filled according to the report information. This will allow admins to save extra time, especially when banning offline players.

Added logging for deleting reports.

Added chat log storage. Admin resource will collect and store most recent chat messages. There is a limit setting in meta.xml which by default is 10 messages per player.
Chat logging is especially useful for reports. For example if a report category "Cursing" is added, regular staff members will be able to see the recent chat log since they do not usually have access to server.log

### DIFF
--- a/[admin]/admin/client/gui/admin_message.lua
+++ b/[admin]/admin/client/gui/admin_message.lua
@@ -9,36 +9,78 @@
 **************************************]]
 
 aViewMessageForm = nil
+local aSuspectInfo = nil
+local viewID
 
 function aViewMessage ( id )
+	if ( aSuspectInfo and viewID and viewID ~= id ) then
+		destroyElement(aSuspectInfo)
+		aSuspectInfo = nil
+	end
+	viewID = id
 	if ( aViewMessageForm == nil ) then
 		local x, y = guiGetScreenSize()
-		aViewMessageForm	= guiCreateWindow ( x / 2 - 150, y / 2 - 125, 300, 250, "", false )
+		aViewMessageForm	= guiCreateWindow ( x / 2 - 150, y / 2 - 129, 300, 258, "", false )
 					   guiCreateLabel ( 0.05, 0.10, 0.30, 0.09, "Category:", true, aViewMessageForm )
-					   guiCreateLabel ( 0.05, 0.18, 0.30, 0.09, "Subject:", true, aViewMessageForm )
-					   guiCreateLabel ( 0.05, 0.26, 0.30, 0.09, "Time:", true, aViewMessageForm )
-					   guiCreateLabel ( 0.05, 0.34, 0.30, 0.09, "By:", true, aViewMessageForm )
+					   guiCreateLabel ( 0.05, 0.18, 0.30, 0.09, "Time:", true, aViewMessageForm )
+					   guiCreateLabel ( 0.05, 0.26, 0.30, 0.09, "By:", true, aViewMessageForm )
+		aViewMessageLabelSuspect = guiCreateLabel ( 0.05, 0.34, 0.30, 0.09, "Suspect:", true, aViewMessageForm )
+		aViewMessageSuspect	= guiCreateButton ( 0.40, 0.34, 0.55, 0.07, "View info", true, aViewMessageForm )
 		aViewMessageCategory	= guiCreateLabel ( 0.40, 0.10, 0.55, 0.09, "", true, aViewMessageForm )
-		aViewMessageSubject	= guiCreateLabel ( 0.40, 0.18, 0.55, 0.09, "", true, aViewMessageForm )
-		aViewMessageTime	= guiCreateLabel ( 0.40, 0.26, 0.55, 0.09, "", true, aViewMessageForm )
-		aViewMessageAuthor	= guiCreateLabel ( 0.40, 0.34, 0.55, 0.09, "", true, aViewMessageForm )
-		aViewMessageText	= guiCreateMemo ( 0.05, 0.41, 0.90, 0.45, "", true, aViewMessageForm )
+		aViewMessageTime	= guiCreateLabel ( 0.40, 0.18, 0.55, 0.09, "", true, aViewMessageForm )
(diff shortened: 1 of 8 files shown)


Dutchman101:
One bug already, the line outputChatBox("removed") constantly outputs ''removed'' in chat to everyone after a report was submitted.

Dutchman101:
I would also suggest not forcing player input with ''reported player missing'' because when server owners set all report categories to isPlayerReport or their initial box selection is such a category, and a player wants to conveniently report something else (you cant expect all players to understand they got to select another category to proceed) **and don't forget** they might be misled by it if wanting to report an offline player and believe they still need a category where ''isPlayerReport'' gets enforced on.

There's no reason to force input and else not let them submit the report, if you can also just strip the suspect details out in the end resulting report. (when they didnt enter a name).
Just don't let it be mandatory I vouch.

Dutchman101:
There also needs to be an escape from ''Select player'' window if player wants to cancel & go a step back, now the only exit is selecting a player. Add a close button

Dezash:
I can't seem to edit admin_server.lua and remove that leftover output. Github gives me 404.

Dezash:
Will this be merged? If something needs to be changed, let me know.

qaisjp:
This will need to be squash merged at some point. @Dezash, please endeavour to use a Git client rather than using the web interface.

Some changes I think need to be made:

* Use of the settings feature, not an XML file for settings
* `Moder` -> `Modder`

[admin]/admin/client/gui/admin_message.lua
```diff
@@ -9,48 +9,99 @@
...
 	if ( _messages[id] ) then
 		guiSetText ( aViewMessageCategory, _messages[id].category )
-		guiSetText ( aViewMessageSubject, _messages[id].subject )
+		guiSetText ( aViewMessageForm, _messages[id].subject )
 		guiSetText ( aViewMessageTime, _messages[id].time )
 		guiSetText ( aViewMessageAuthor, _messages[id].author )
 		guiSetText ( aViewMessageText, _messages[id].text )
 		guiSetVisible ( aViewMessageForm, true )
+		if ( _messages[id].suspect ) then
+			guiSetVisible ( aViewMessageLabelSuspect, true )
```
Comment (qaisjp) [inline [admin]/admin/client/gui/admin_message.lua]: Use `guiSetVisible( gui, _messages[id].suspect )` instead.

[admin]/admin/server/admin_serverprefs.lua
```diff
@@ -80,12 +81,12 @@ end
 --------------------------------------------------------------------------------
 -- Coroutines
 --------------------------------------------------------------------------------
--- Make sure errors inside coroutines get printed somewhere
```
Comment (qaisjp) [inline [admin]/admin/server/admin_serverprefs.lua]: What are these changes? Line endings?

[admin]/admin/server/admin_server.lua
```diff
@@ -1376,6 +1455,10 @@ addEventHandler ( "aMessage", _root, function ( action, data )
 	end
 end )
 
+addCommandHandler ( "report", function ( player )
+	triggerClientEvent ( player, "onReport", player, aReportCategories )
```
Comment (qaisjp) [inline [admin]/admin/server/admin_server.lua]: What is the purpose of this change if just to trigger a client event? What happens if a user runs `report` before this resource has loaded clientside?

[admin]/admin/server/admin_server.lua
```diff
@@ -97,58 +99,83 @@ addEventHandler ( "onResourceStart", _root, function ( resource )
...
+				aReportCategories[i] = {}
+				local isReport = xmlNodeGetAttribute ( child, "isPlayerReport" )
+				if ( isReport and isReport == "true" ) then
+					aReportCategories[i].playerReport = true
+				end
+				aReportCategories[i].subject = xmlNodeGetValue ( child )
+			end
 		end
-		xmlUnloadFile ( node )
+	xmlUnloadFile ( node )
```
Comment (qaisjp) [inline [admin]/admin/server/admin_server.lua]: Is this indentation correct? If not, please fix.

[admin]/admin/server/admin_serverprefs.lua
Comment (Dezash) [inline [admin]/admin/server/admin_serverprefs.lua]: I don't know what that is.

[admin]/admin/server/admin_server.lua
Comment (Dezash) [inline [admin]/admin/server/admin_server.lua]: That's because aReportCategories is server-sided. If that's a problem, I could make aReportCategories be sent when client resource starts and thus make the command client-side.

[admin]/admin/client/gui/admin_message.lua
Comment (Dezash) [inline [admin]/admin/client/gui/admin_message.lua]: _messages[id].suspect is a table.

Dezash:
By settings feature, do you mean the set/get functions?

[admin]/admin/client/gui/admin_message.lua
Comment (qaisjp) [inline [admin]/admin/client/gui/admin_message.lua]: So "suspect" would be `nil`. Would `guiSetVisible( gui, _messages[id].suspect ~= nil )` work?

[admin]/admin/server/admin_server.lua
Comment (qaisjp) [inline [admin]/admin/server/admin_server.lua]: That would be a better solution.

qaisjp:
@Dezash yes, `set`/`get` functions.

Bonus points for supporting live updates.

Dezash:
Settings support live updates and the report command was moved back to client-side. I also added a category "Spammer".

Dutchman101:
You would say with so many changes SOMETHING has to break, but after thoroughly testing it seems to be stable.
Furthermore all recommendations from codereview were implemented, without new remarks. Merging.

PR #62, opened 2017-01-06
[editor] Stop creeping out mappers

lopezloo:
(disable CJ breathing and other related sounds in edit mode)
These sounds still works in test mode (thanks to resetMapInfo
somewhere).

https://forum.mtasa.com/topic/92093-bug-myth-breathing-sound/


qaisjp:
The original functionality is an intended feature.

PR #63, opened 2017-01-11
Some bugfixes (+ #64)

emre1702:
1. You could eject yourself immediately by pressing F (enter exit vehicle) and mouse2 (driveby) - without animations.
That happened because of setPedDoingGangDriveby.
2. If you used a weapon, which isn't allowed for the driver, on another seat, you couldn't switch weapons as driver.
3. You could use not allowed weapons as a drive by using the weapon on another seat and go back as driver
4. Switching weapons let you shoot fast with slow weapons like Deagle
5. New function to replace the hitbox while using driveby on a bike with the hitbox of an attached ped - only working for clientsided damagesystem with attacker triggering the damage
6. Better performance by localizing the variables

### DIFF
--- a/[gameplay]/realdriveby/driveby_client.lua
+++ b/[gameplay]/realdriveby/driveby_client.lua
@@ -97,7 +97,15 @@ addEventHandler("doSendDriveBySettings",localPlayer, function ( newSettings, the
 		peds[i] = v 
 		pedowner[v] = i
 	end
-end )
+	if eject_on_driveby_fix then
+		--[[ Block driveby when exiting vehicle: 
+  		setPedDoingGangDriveby ejects the player if he is trying to get out of the vehicle (without animations ) ]]
+		addEventHandler ( "onClientVehicleStartExit", root, function ( player )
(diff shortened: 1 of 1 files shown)


Dutchman101:
''You could eject yourself immediately by pressing F (enter exit vehicle) and mouse2 (driveby) - without animations.''

I got my doubts with this, players on several gamemode types became adapted to this on a sort of ''glitch'' because they can circumvent the known MTA bug with it that once you pressed 'F' (exit) it can stall long until your ped jumps out of the vehicle (until its decelerated enough) for example you cannot just exit when vehicle is (falling) in the air or driving on high speeds, by triggering exit and when it doesnt respond immediately, pressing the Right mouse button to initiate driveby, which in this case will eject the player out of the blocked position.

Dutchman101:
Too many players adapted to that method of fixing such a situation that I want to suggest only putting it as toggle'able setting, and then keeping the default value as disabled.
Server owners may then opt for enabling the fix.

emre1702:
Added the setting to do that

PR #66, opened 2017-01-11
realdriveby fixes

emre1702:
Here are some realdriveby fixes.
This resource got some bugs which were really annoying.

I hope this time I did it right.

### DIFF
--- a/[gameplay]/realdriveby/driveby_client.lua
+++ b/[gameplay]/realdriveby/driveby_client.lua
@@ -21,6 +21,7 @@ local function setupDriveby( player, seat )
 		toggleDriveby()
 	end
 	exitingVehicle = false
+	lastSlot = 0
 end
 addEventHandler( "onClientPlayerVehicleEnter", localPlayer, setupDriveby )
 
--- a/[gameplay]/realdriveby/driveby_client.lua
+++ b/[gameplay]/realdriveby/driveby_client.lua
@@ -261,10 +261,8 @@ function limitedKeyPress (key,keyState,speed)
 		limiterTimer = setTimer ( pressKey,speed, 0, "vehicle_fire" )
 	else
 		shooting = false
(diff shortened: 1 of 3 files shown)


Dutchman101:
I suggested to default the setting for the eject fix on False at your initial PR, and I stay with that; you made it default to true. I explained there why I feel opt-in by server admins is desired in my opinion.

qaisjp:
setGlitchEnabled is opt-in for keeping glitches. The default functionality
should always be fixed. Server owners should read the fine print when
updating resources.

emre1702:
Don't know how it is in other countries, but in every german roleplay server (I played) the using of this bug gets you a ban. Played at every big server (since 2013) and it was always the same - they call it "bugusing".
So I would rather let default at true.

[gameplay]/realdriveby/driveby_client.lua
```diff
@@ -54,6 +58,25 @@ addEventHandler("doSendDriveBySettings",localPlayer,
...
+		settings.driverallowed = {}
+		if settings.driver[1] then
+			for i=1, #settings.driver do
+				settings.driverallowed[settings.driver[i]] = true
+			end
+		end
+		settings.passengerallowed = {}
+		if settings.passenger[1] then
+			for i=1, #settings.passenger do
+				settings.passengerallowed[settings.driver[i]] = true
```
Comment (Jusonex) [inline [gameplay]/realdriveby/driveby_client.lua]: Throws an error on startup (first start after fetching your PR): `ERROR: realdriveby\realdriveby_client.lua:70: table index is nil`

Comment (Jusonex) [inline [gameplay]/realdriveby/driveby_client.lua]: For the rest, it looks good to me.
I also agree with _qaisjp_ that the default behavior should be fixed.

emre1702:
Used the wrong table, fixed

Dutchman101:
Tested thorougly and stable, and no one made further codereview remarks, so merging.

PR #67, opened 2017-02-26
Remove unnecessary showCursor function

Dezash:
The cursor is already handled by admin panel and the extra showCursor function in colorpicker just makes it difficult for other resources to use it.


Jusonex:
Thanks for the patch.
However, since the color picker is meant to be an independant component, it's okay to have the redundant _showCursor_ here.

PR #68, opened 2017-04-04
Fix admin resource bug when delete resource

AboShanab:
It's fix bug visible for me sometimes when delete resource..


qaisjp:
What if the resource was also in the `stopping` or `starting` state?

AboShanab:
I don't have any idea if that bug will effect with those statuses, but anyway that pull request will fix the bug what I see it in debugscript command.

AboShanab:
I will remove that pull request, Because the code not useful as well.

PR #70, opened 2017-04-30
Update for IPB resource

Necktrox:
**Changes**
- Dynamic window size, which depends on screen resolution
- Added support for client performance statistics
- Separated code base into multiple files
- Improved resource settings config
- Auto-sized statistic columns

## **1920x1080**
ipb

## **1280x720**
ipb

## **800x600**
ipb

## **640x480**
ipb


qaisjp:
It would've been best to do the separation separately because right now we can't see what changes have been made.

Necktrox:
I am not sure if that would've helped in any way, because most of the code has been changed.

Dutchman101:
quite a few files with tabulation issues (empty, extra space on line 1)

Necktrox:
@Dutchman101 I keep the first and last line empty as my own design choice. The last commit should fix the remaining indentation mistakes.

Dutchman101:
100% stable

qaisjp:
Thank you!

PR #72, opened 2017-06-11
Fix glitch: fast jetpack flying with gravity

Dutchman101:
This PR fixes the glitch that if you set gravity to ''strong'' while using jetpack, you'll be able to fly around fast as a rocket. Flying faster with jetpack than intended can give unfair advantage, infact this glitch makes it look like you're using a speedhack.

It will fix by not allowing jetpack usage on custom gravity.


Dutchman101:
PR messed up due to unrelated commit mixing

PR #73, opened 2017-07-02
overhaul/modernisation of play gamemode

Dutchman101:
This PR contains several improvements to play mode;

1) made vehicles static so they're always available to take out for a drive (previously, only after the server/resource started they would last, as pretty quickly the .map vehicles were torn apart and moved out, making the purpose of having a whole map file short-lived.

2) fixed players spawning into eachother (as exact same coords were used for all spawns) that often resulted in being unable to move (players stuck into eachother) and with that removed dependency on spawnmanager

3) replicated everything from the .map into broph.lua, meaning the spawnpoints are no longer stored in the .map, becoming much more efficient. The spawned vehicles are also done by broph.lua now, as it was required for the static vehicles feature.

4) OOP

I think it's more player-friendly as having vehicles in spawns was for ease of use and for example use by non-completed download players & those who don't yet know how to spawn vehicles (not uncommon in freeroam servers) or are just looking for easy vehicles to grab and be gone quicker.
Anyways, as this would be the logical clarification for having map vehicles initially by the .map as present, it's a bummer this was short-lived as explained in point 1) which is now worked around.

Clarification: vehicles being static means frozen in their position, unable to move UNTIL a player enters it and drives off, to prevent deathmatching of vehicles when not yet used they are undestroyable until actually taken by someone, hence called static. The damageproof property and frozen state will release as soon someone takes the vehicle, after which the gap will be filled up by the vehicle respawning in 1 minute time.
The player will be allowed to continue driving the matching car though, until it eventually expires/destroys after they abandon it.

In broph.lua, the exact positioning and exact car models from .map were taken into regard while spawning everything to look equally.


[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,48 +1,179 @@
...
+local playerVehicles = {}
+
+local function spawn(player)
 
-function spawn(player)
-	if not isElement(player) then return end
-	if get("spawnreset") == "onSpawn" then
-		spawnpoint = getRandomSpawnPoint()
+	if player and isElement(player) and getElementType(player) == "player" then
```
Comment (qaisjp) [inline [gamemodes]/[play]/play/broph.lua]: Is this required?

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,48 +1,179 @@
...
+		local x,y,z,r = unpack(spawnpoints[math.random(1,#spawnpoints)])
+		spawnPlayer(player,x+math.random(-3,3),y+math.random(-3,3),z,r,validSkins[math.random(1,#validSkins)],0,0)
+		fadeCamera(player, true)
+		setCameraTarget(player, player)
+		showChat(player, true)
 	end
-	exports.spawnmanager:spawnPlayerAtSpawnpoint(player,spawnpoint,false)
-	repeat until setElementModel(player,math.random(312))
```
Comment (qaisjp) [inline [gamemodes]/[play]/play/broph.lua]: Removing this was a good idea. If a player left the server before this line was reached then this would result in an infinite loop.

The below snippet would have worked too, but is no longer necessary because of the use of spawnPlayer above.

```lua
repeat until !isElement(player) or setElementModel(...)
```

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,48 +1,179 @@
...
 	end
-)
+	
+	playerVehicles[source] = nil
+
+end
+
+local function onEnter(player)
+
+	local t = get("vehicleRespawnTime")
```
Comment (qaisjp) [inline [gamemodes]/[play]/play/broph.lua]: What happens if this is not a number? Perhaps validation should be done onResourceStart?

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,48 +1,179 @@
...
-addEventHandler("onResourceStart", resourceRoot,
-	function()
-		spawnpoint = getRandomSpawnPoint()
-		resetMapInfo()
-		for i,player in ipairs(getElementsByType("player")) do
-			spawn(player)
-		end
-	end
-)
+local vehiclesToSpawn =
```
Comment (qaisjp) [inline [gamemodes]/[play]/play/broph.lua]: Is a table instead of a map required? [getResourceMapRootElement()](https://wiki.multitheftauto.com/wiki/GetResourceMapRootElement) could be used to freeze all the vehicles. And then a single handler could be applied to that root element for enter/leave.

The respawn functions could be used as well: respawnVehicle, setVehicleIdleRespawnDelay, setVehicleRespawnDelay, setVehicleRespawnPosition, etc

Comment (Dutchman101) [inline [gamemodes]/[play]/play/broph.lua]: About the respawn funcs already; thinking about if lot of players would take those vehicles from spawn without never leaving them, just having fun with them
It MAY cause lack of vehicles, the destroy/expire timer takes care of those no longer used while the spawning back is now independent process.

[gamemodes]/[play]/play/broph.lua
Comment (Dutchman101) [inline [gamemodes]/[play]/play/broph.lua]: maybe checking element type is too much, but there must be a check that the element variable exists and it is a element because it caused debug warnings;
Bad argument @ 'spawnPlayer' [Expected player at argument 1]
Bad argument @ 'fadeCamera' [Expected element at argument 1]
Bad argument @ 'showChat' [Expected element at argument 1]

[gamemodes]/[play]/play/broph.lua
Comment (Dutchman101) [inline [gamemodes]/[play]/play/broph.lua]: solved this in latest added commit (bbcc149)

[gamemodes]/[play]/play/broph.lua
Comment (Dutchman101) [inline [gamemodes]/[play]/play/broph.lua]: Removed the excessive elementype check, letting the others remain for given reason.

Dutchman101:
No further objections or feedback after fixes, so merged.

PR #74, opened 2017-07-15
getControlState is deprecated

Ares-C:


Jusonex:
Thanks for the patch.
Since old clients are still around, we'll delay merging the patch until the next release.

PR #76, opened 2017-08-05
Support hydra and hunter kills > driven vehicles in killfeed

Dutchman101:
This will make hydra/hunter kills on vehicles show up in killfeed. Previously when you shot a player in vehicle directly with the Hydra/hunter rockets or hunter machinegun, it would not detect you as killer and show ''blown'' only.
The method used is the only way to detect these kills, although I've never seen this ''guessing'' being inaccurate so far. Tested for months.

### DIFF
--- a/[gameplay]/killmessages/killmessages_client.lua
+++ b/[gameplay]/killmessages/killmessages_client.lua
@@ -1,4 +1,33 @@
-﻿addEvent ("onClientPlayerKillMessage",true)
+local specialVehicleDamagerModels =
+{
+	[520] = true,
+	[425] = true,
+}
+
+function onAirstrikeDamaged(att, wep, loss)
+	local localVeh = getPedOccupiedVehicle(localPlayer)
+	if att and getElementType(att) == "player" and att ~= localPlayer and localVeh and source == localVeh and loss > 0 then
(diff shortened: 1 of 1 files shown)


[gameplay]/killmessages/killmessages_client.lua
```diff
@@ -1,4 +1,33 @@
-﻿addEvent ("onClientPlayerKillMessage",true)
+local specialVehicleDamagerModels =
+{
+	[520] = true,
+	[425] = true,
+}
+
+function onAirstrikeDamaged(att, wep, loss)
+	local localVeh = getPedOccupiedVehicle(localPlayer)
+	if att and getElementType(att) == "player" and att ~= localPlayer and localVeh and source == localVeh and loss > 0 then
```
Comment (qaisjp) [inline [gameplay]/killmessages/killmessages_client.lua]: Try to use

```
if not a then
     return
end
```

instead of

```
if a then
    ...
end
```

-----

Also, this is a bit long, use paranthesisesee too and split over multiple lines:

```
if (not att) or (getElementType(att) ~= "player") or
   (att == localPlayer) or (not localVeh) or
   (source ~= localVeh) or (loss <= 0)
then
     return
end
````

same applies to isPedInVehicle below

Dutchman101:
It was an old patch lying around, I had to add another commit but I just discovered serverside was based off outdated killmessages revision.
So was a bit too quick opening the PR, delayed and might open a proper one soon.

qaisjp:
ok

PR #84, opened 2017-10-23
[editor] Added 'sirens' option and fixed 'plate'

MIKI785:
Added a new 'sirens' option that toggles sirens on vehicles that have them.
Also, since we are now able to change vehicle's number plates after it's created, I added this functionality (the plates now change as the property is changed).

### DIFF
--- a/[editor]/edf/edf.edf
+++ b/[editor]/edf/edf.edf
@@ -38,6 +38,7 @@
 		<data name="paintjob" type="selection:0,1,2,3" description="vehicle paintjob" required="false" />
 		<data name="upgrades" type="vehicleupgrades" description="installed vehicle upgrades" required="false" />
 		<data name="plate" type="plate" description="number plate" required="false" />
+		<data name="sirens" type="selection:Off,On" description="Vehicle sirens" required="false" default="Off"/>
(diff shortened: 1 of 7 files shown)


[editor]/edf/edf.edf
```diff
@@ -38,6 +38,7 @@
 		<data name="paintjob" type="selection:0,1,2,3" description="vehicle paintjob" required="false" />
 		<data name="upgrades" type="vehicleupgrades" description="installed vehicle upgrades" required="false" />
 		<data name="plate" type="plate" description="number plate" required="false" />
+		<data name="sirens" type="selection:Off,On" description="Vehicle sirens" required="false" default="Off"/>
```
Comment (qaisjp) [inline [editor]/edf/edf.edf]: Thank you very much for you contribution!

The code should use "false,true" rather than "Off/On". See the use of "doublesided" above: https://github.com/multitheftauto/mtasa-resources/pull/84/files#diff-cfed86a3446d57ccd8ceb88c535cb3e4R9

Comment (MIKI785) [reply [editor]/edf/edf.edf]: That's because that is the value that gets saved into the .map file, right? I've changed it.

Dutchman101:
qaisjp already checked conformity, I tested the changes and it's all stable, so merging.

PR #86, opened 2017-11-18
EDF fixes

jlillis:
This pull request:

- Fixes a bug where custom EDF elements will have their representative elements created with in the wrong interior
- Uses [getValidPedModels](https://wiki.multitheftauto.com/wiki/GetValidPedModels) to validate ped skin IDs in datatypes.lua
- Fixes an annoying issue where an EDF elements with required attributes that don't have a default value would always be created at 0,0,0 and detached from the cursor. This was most noticeable when trying to create an "Interior warp return" element from the interiors EDF.
- Trims some trailing whitespace (this repo could use an [.editorconfig](http://editorconfig.org/) file)


Dutchman101:
Trailing spaces have been addressed by earlier global trimming.
Because of some unclear conflicts I just commited your changes; https://github.com/multitheftauto/mtasa-resources/commit/df92f9b9827586506b6e3a29175eee7b079534cf

Thanks for the patch @jlillis and I apologize for the delay in review.

PR #88, opened 2018-01-20
[admin] Added logging for ACL changes, unban etc

Dutchman101:
When admins change something in ACL (add user, right, resource, new ACL, delete nodes, or rights) it will now log who did that and specify the exact right/element that was modified and in which way. Previously, a nondescriptive (hardcoded server) message just stated something had changed only, so for server owners, (malicious) actions could hardly be traced back to a specific admin.

Also added: more extensive logging for unbanning and banning offline players, for example it will now specify the serial/IP which gets unbanned and by who, instead of just stating a ban was removed.

### DIFF
--- a/[admin]/admin/conf/messages.xml
+++ b/[admin]/admin/conf/messages.xml
@@ -247,22 +247,7 @@
     </group>
   </server>
   <bans>
-    <group action="unbanip" r="225" g="170" b="90">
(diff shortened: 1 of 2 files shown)


Dutchman101:
Also a small bugfix for kicking players with Anonymous admin toggled, is included.

PR #90, opened 2018-03-08
Update deathmessages.lua

Pirulax:
Maybe its faster this way, we dont need that 'root' in local anyway.


Dezash:
It isn't faster.

PR #94, opened 2018-04-07
mantis-9838: [race] mapmanager is not included in the meta.xml yet still required in code, throws errors in server console

patrikjuvonen:
`[gamemodes]/[race]/race/racevoting_server.lua:442` would throw server-side errors if the `mapmanager` resource was not running.

```
ERROR: [gamemodes]\[race]\race\racevoting_server.lua:442: exports: Call to non-running server resource (mapmanager) [string "?"]
```

Bug report at https://bugs.mtasa.com/view.php?id=9838

### DIFF
--- a/[gamemodes]/[race]/race/_common.lua
+++ b/[gamemodes]/[race]/race/_common.lua
@@ -323,6 +323,11 @@ function matrixToRotations( matrix )
 
 	return rx, ry, rz
 end
+
+-- Check if the race resource is ready to be started
+function isRaceResourceReady()
+	return getResourceState(getResourceFromName('mapmanager')) == 'running'
+end
 ---------------------------------------------------------------------------
 
 
--- a/[gamemodes]/[race]/race/_common.lua
+++ b/[gamemodes]/[race]/race/_common.lua
@@ -19,25 +19,25 @@ function getScriptLocation()	return isServer() and "Server" or "Client"	end
 -- Math extentions
 ---------------------------------------------------------------------------
 function math.lerp(from,to,alpha)
-    return from + (to-from) * alpha
+	return from + (to-from) * alpha
 end
 
 function math.clamp(low,value,high)
-    return math.max(low,math.min(value,high))
+	return math.max(low,math.min(value,high))
 end
 
 function math.wrap(low,value,high)
-    while value > high do
-        value = value - (high-low)
-    end
-    while value < low do
-        value = value + (high-low)
-    end
-    return value
+	while value > high do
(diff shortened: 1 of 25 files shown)


patrikjuvonen:
Seems we need to figure out a different approach to the problem. While including it into the meta.xml removes the error, it will not start the map as race depends on onGamemodeStart. Silly, I'll look into what small things we can do to keep it all in one package.

patrikjuvonen:
Implemented `isRaceResourceReady` shared function into `_common.lua`. This is used in on(Client)ResourceStart handlers. This is probably not the way to go, though. Sounds _hacky_, but the baseline is that `mapmanager` is required and for those who do not know how to set up race gamemode, it seems like a waste of time for `race` not to handle itself, as it can handle about everything else itself.
So, awaiting for feedback, really.

qaisjp:
> While including it into the meta.xml removes the error, it will not start the map since race depends on onGamemodeStart.

Why won't onGamemodeStart trigger if mapmanager is an include?

patrikjuvonen:
@qaisjp Apparently because it's triggered too early in the process.

qaisjp:
OK... what do you think we can add to the MTA core to help make this fix cleaner? I am happy to approve this now but if you think we should have some sort of attribute on `<include>` to control the timing of the resource then we can look into that.

I don't know how complex MTA treats this "dependency" system - I imagine it's pretty primitive, but I could be wrong.

patrikjuvonen:
Well, I believe we should be able to specify if we want an included resource prepared before the parent resource is ready. Because that would fix this problem - we could start `mapmanager` _before_ `race` instead of after, as it appears here, or at least it doesn't trigger the event as it should...

Woovie:
> Well, I believe we should be able to specify if we want an included resource prepared and started before the parent resource is ready (as an attribute in the include tag).
We should look at how systemd handles this inside their unit files, as I think this is a good implementation.

https://www.freedesktop.org/software/systemd/man/systemd.unit.html

Specifically https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Mapping%20of%20unit%20properties%20to%20their%20inverses

We could implement requires and wants in meta. This would allow you to start required or optional resources from within meta. As well, we can implement before and after.

https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Before=

patrikjuvonen:
I'm going to make an issue about this and close this PR soon.

jlillis:
I'm going to close this for now, as per the previous comment in April. Feel free to reopen this in the future if you like.

PR #109, opened 2018-07-24
Fixed cannot resume dead coroutine.

CodyJL:
Added verification to check if the coroutine is currently 'Alive'


CodyJL:
New commit pushed.

PR #128, opened 2018-09-26
admin2: Removed variables which are unnecessary

xLuxy:
This PR removes some variables which are not required / unnecessary.

`_local: Only used once in onResourceStart`
`_root: We're having a predefined variable called "root" already defined / declared`

- Luxy.c


[admin]/admin2/server/admin_server.lua
```diff
@@ -7,7 +7,7 @@
 *	Original File by lil_Toady
 *
 **************************************]]
-_root = getRootElement()
+root = getRootElement()
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: oops

Comment (xLuxy) [reply [admin]/admin2/server/admin_server.lua]: root is already declared (predefined variable of MTA)

Comment (qaisjp) [reply [admin]/admin2/server/admin_server.lua]: Exactly, shouldn't it be deleted?

PR #129, opened 2018-09-26
admin2: renamed conf/settings.db to admin.db

xLuxy:
Renames current conf\settings.db to admin.db. The reason for this change is for further implementations such as whowas, network tab, logging and maybe more :)

- Luxy.c

### DIFF
--- a/[admin]/admin2/server/admin_database.lua
+++ b/[admin]/admin2/server/admin_database.lua
@@ -9,7 +9,7 @@
 **************************************]]
 
 -- ensure old database file gets renamed
-if(fileExists("conf/settings.db")) then
+if (fileExists("conf/settings.db")) then
     fileRename("conf/settings.db", "admin.db")
 end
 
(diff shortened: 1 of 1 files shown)


[admin]/admin2/server/admin_database.lua
```diff
@@ -9,7 +9,9 @@
 **************************************]]
 
 -- ensure old database file gets renamed
-fileRename("conf\\settings.db", "admin.db")
+if(fileExists("conf\\settings.db")) then
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_database.lua]: Enclosing parenthesis not necessary. Should be space after `if`.

[admin]/admin2/server/admin_database.lua
```diff
@@ -9,7 +9,7 @@
 **************************************]]
 
 -- ensure old database file gets renamed
-if(fileExists("conf/settings.db")) then
+if (fileExists("conf/settings.db")) then
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_database.lua]: Extra parenthesis still exists

Comment (qaisjp) [reply [admin]/admin2/server/admin_database.lua]: Preferred syntax is no brackets but considering it is used everywhere else we can fix that line when we fix the rest

PR #143, opened 2019-01-20
race: Fix #139

### LINKED ISSUE #139: race: Next map vote Lua error if there are no maps on the server
race

**Describe the bug**
`race\racevoting_server.lua:217: attempt to index global 'errorCode' (a nil value)`

**Code**
https://github.com/multitheftauto/mtasa-resources/blob/17da6d7e3bc0e88014b579756e31f1426e9bee8f/%5Bgamemodes%5D/%5Brace%5D/race/racevoting_server.lua#L217

ciber96:
Fixes #139 
Implement error codes for race and reimplement suppressed error codes in votemanager.


ciber96:
Done

[gamemodes]/[race]/race/race_errorcodes.lua
```diff
@@ -1,20 +1,30 @@
...
-}
\ No newline at end of file
+errorCode = {}
+
+local function loadAdditionalErrorCodes()
+	-- Load error codes from votemanager resource if it's running
+	local votemanagerResource = getResourceFromName("votemanager")
+	if (votemanagerResource) then
+		if (getResourceState(votemanagerResource) == "running") then
+			local votemanagerErrorCodes = exports.votemanager:getErrorCodes()
```
Comment (qaisjp) [inline [gamemodes]/[race]/race/race_errorcodes.lua]: Why not just do a direct assignment?

Comment (patrikjuvonen) [reply [gamemodes]/[race]/race/race_errorcodes.lua]: I had extensibility in mind and having the ability to override error codes in race.

Comment (qaisjp) [reply [gamemodes]/[race]/race/race_errorcodes.lua]: ok

[gamemodes]/[race]/race/race_errorcodes.lua
```diff
@@ -0,0 +1,30 @@
...
+			if (votemanagerErrorCodes) and (type(votemanagerErrorCodes) == "table") then
+				for key, code in pairs(votemanagerErrorCodes) do
+					-- If the error code is not defined here already, then set it
+					if (errorCode[key] == nil) then
```
Comment (qaisjp) [inline [gamemodes]/[race]/race/race_errorcodes.lua]: I think that having the ability for race to override keys *might* actually cause confusion in the future. I think instead of ignoring already defined codes, we should error out: `assert(errorCode[key] == nil)`

What do you think?

Comment (patrikjuvonen) [reply [gamemodes]/[race]/race/race_errorcodes.lua]: Sounds fair enough, I've changed this now.

Woovie:
Is this ready for merge?

patrikjuvonen:
Thanks for the pull request and sorry it took a long time :)

PR #153, opened 2019-06-13
admin2: Fix #152

### LINKED ISSUE #152: admin2: misbehaviour if you perform user operations without mouse-selecting them

admin2

**Describe the bug**

Some details do not update if you select yourself this way.

**To Reproduce**

1. Restart `admin2`
2. Press `p` to load the admin panel
3. Press the DOWN arrow key to select yourself
1. Note how `ip`, `serial`, `country`, `account` and `groups` are all UNSET
4. Press `mute`, and notice how the mute button does not now say "Unmute"
5. Press `mute` again, and it tries to mute you _again_

**Expected behavior**

1. Select yourself in the sidebar using your mouse
1. Note how `ip`, `serial`, `country`, `account` and `groups` are all set correctly now
1. Note how the `Mute` button now says `Unmute` (and works correctly now)

**Screenshots**

**Version**

**Additional context**

image
admin

FileEX:
Fixes #152 
Now if we use DOWN or UP arrow key, informations like freeze,mute,ip,serial etc will be update/set

### DIFF
--- a/[admin]/admin2/client/main/admin_players.lua
+++ b/[admin]/admin2/client/main/admin_players.lua
@@ -375,26 +375,6 @@ function aPlayersTab.onClientClick(button)
             local player = getSelectedPlayer()
             if (player) then
                 aPlayersTab.onRefresh()
-                sync(SYNC_PLAYER, player)
-                guiSetText(aPlayersTab.IP, "IP: " .. aPlayers[player].ip)
-                guiSetText(aPlayersTab.Serial, "Serial: " .. (aPlayers[player].serial or "Unknown"))
-                guiSetText(aPlayersTab.Country, "Country: " .. (aPlayers[player].countryname or "Unknown"))
-                guiSetText(aPlayersTab.Account, "Account: " .. (aPlayers[player]["account"] or "guest"))
-                guiSetText(aPlayersTab.Groups, "Groups: " .. (aPlayers[player]["groups"] or "None"))
-                if (aPlayers[player].country and string.lower(tostring(aPlayers[player].country)) ~= "zz") then
-                    local x, y = guiGetPosition(aPlayersTab.Country, false)
-                    local width = guiLabelGetTextExtent(aPlayersTab.Country)
-                    guiSetPosition(aPlayersTab.Flag, x + width + 3, y + 4, false)
-                    guiSetVisible(
-                        aPlayersTab.Flag,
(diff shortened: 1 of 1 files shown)


FileEX:
Done

[admin]/admin2/client/main/admin_players.lua
```diff
@@ -567,6 +552,27 @@ function aPlayersTab.onRefresh()
...
     end
+    
+    sync(SYNC_PLAYER, player)
+    guiSetText(aPlayersTab.IP, "IP: " .. aPlayers[player].ip)
+    guiSetText(aPlayersTab.Serial, "Serial: " .. (aPlayers[player].serial or "Unknown"))
+    guiSetText(aPlayersTab.Country, "Country: " .. (aPlayers[player].countryname or "Unknown"))
+    guiSetText(aPlayersTab.Account, "Account: " .. (aPlayers[player]["account"] or "guest"))
+    guiSetText(aPlayersTab.Groups, "Groups: " .. (aPlayers[player]["groups"] or "None"))
+    if (aPlayers[player].country and string.lower(tostring(aPlayers[player].country)) ~= "zz") then
+       local x, y = guiGetPosition(aPlayersTab.Country, false)
```
Comment (patrikjuvonen) [inline [admin]/admin2/client/main/admin_players.lua]: The indentation here is 3 spaces instead of 4.

[admin]/admin2/client/main/admin_players.lua
```diff
@@ -622,7 +628,7 @@ function aPlayersTab.onRefresh()
         guiSetText(aPlayersTab.Vehicle, "Vehicle: Foot")
         guiSetText(aPlayersTab.VehicleHealth, "Vehicle Health: 0%")
     end
-
+
```
Comment (patrikjuvonen) [inline [admin]/admin2/client/main/admin_players.lua]: Unwanted whitespace added.

[admin]/admin2/client/main/admin_players.lua
```diff
@@ -479,6 +459,11 @@ function aPlayersTab.onPlayerListScroll(key, state, inc)
     else
         guiGridListSetSelectedItem(aPlayersTab.PlayerList, next, 1)
     end
+
+    -- if we finnally have selected any item
```
Comment (qaisjp) [inline [admin]/admin2/client/main/admin_players.lua]:
```suggestion
    -- If we finally have selected an item
```

Is this `if` check needed? It seems that by the time we get to this line, we will always have selected an item.

PR #155, opened 2019-06-13
editor: Fix #142

### LINKED ISSUE #142: editor_main: save doesn't save element dimension
editor_main

**Describe the bug**
Saving map doesn't save element dimensions.

**To Reproduce**
1. Spawn an object, vehicle or marker
2. Set its dimension to something other than 0
3. Save the map
4. Open the map file
5. `dimension` attribute's value is `0` for all elements

**Expected behavior**
Save elements' set dimension.

**Additional context**
[Related comment](https://github.com/multitheftauto/mtasa-blue/issues/640#issuecomment-428652697) by @botder 

I'm currently working on a PR to close this issue.

FileEX:
Fixes #142 
Now dimension is save correctly.


patrikjuvonen (dismissed):
Unfortunately this fix has the side-effect of elements set into other dimensions are forever hidden as map (editor) dimension cannot be changed.

This fix should instead only affect the save but not real-time.

FileEX:
Fixed!

jlillis:
I've confirmed this PR works as intended and resolves #142 and #190

PR #158, opened 2019-07-18
Remove unused code

Dezash:
Removes empty and unused functions from the admin resource.
admin_editor.lua was deleted because it doesn't contain any code whatsoever.


qaisjp:
Thanks!

Issue #116, opened 2018-07-29
admin2: whowas command


qaisjp:
**Is your feature request related to a problem? Please describe.**
In https://github.com/multitheftauto/mtasa-blue/commit/0cffa9215f4601e42b2dfbc254040aeda3caa8f0 the `whowas` command was removed. I can't use it anymore 😢 

**Describe the solution you'd like**
Please implement a new `whowas` command.

**Describe alternatives you've considered**
n/a

**Additional context**
A GUI for this would also be cool.

MIKI785:
How exactly is this supposed to work? Because the old command returns only one result for some players, then multiple different result for others and the same result multiple times for the rest.

PR #161, opened 2019-07-19
Added whowas command

PlatinMTA:
This should satisfice #116. This commit should add the command /whowas. The data is stored on the admin2 database. The table "whowas" on the database will store: playerName, lastIP, lastSerial and timeStamp.

Every time a player joins or quits the server the database will update. There could be many rows with the same playerName but it will certanly have a different serial. When using /whowas you will get all the information on the console.

Someone should check this commit before mergin it tho, I didn't find any problems but I also didn't test that much.


Dezash:
This creates a lot of unnecessary database queries that could affect server's performance. You could optimize the code significantly by caching data in a table and saving it to the database when the resource is stopped.

PlatinMTA:
> This creates a lot of unnecessary database queries that could affect server's performance. You could optimize the code significantly by caching data in a table and saving it to the database when the resource is stopped.

Yes, you are right about the queries, but I don't really know how much of a performance issue this could cause and if having all that data on a table is even worth it. Just imagine having 10k rows on a single table and trying to search through it. Maybe I'm wrong and it goes ultra fast, I doubt it tho.

What I could do is have a recent user connection/disconnection table and save it when it reaches certain ammount of rows or when the resource stops, and then when we run the command it will go first through the table and if it doesn't find anything then it goes to the database. That should be a good solution.

[admin]/admin2/server/admin_server.lua
```diff
@@ -76,6 +93,13 @@ addEventHandler(
             end
         end
         setPedGravity(source, getGravity())
+
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: This would benefit from being in its own function instead of copied several times

[admin]/admin2/server/admin_server.lua
```diff
@@ -76,6 +93,13 @@ addEventHandler(
             end
         end
         setPedGravity(source, getGravity())
+	
+	-- whowas stuff
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: This would benefit from being in its own function instead of copied several times

Comment (PlatinMTA) [reply [admin]/admin2/server/admin_server.lua]: Any problem if I put it just before the (addEventHandler, "onResourceRoot"... ?

PlatinMTA:
> Sounds good, you could also add settings to enable/disable the feature and to set the maximum amount of entries that it saves.

That's a good idea. I would do it tomorrow tho, it's like 5am here.

PlatinMTA:
Ok I just updated and made the table thing. You have three new settings now:

whowassave -> which selects which method you want to use for storing info
whowasmax -> which only works if whowassave is "both", it says which would be the maximum amount of rows into aWhowas
whowasprint -> which will set where should we print the info (if on console or chat).

Because saving into a table won't query the info on the database unless the table doesn't have the info we want, sometimes all the info won't be displayed. For that reason I choosed to set the default to "onlydb" on whowassave.

Other than that it should work just fine. You can search by name or by serial, depending of which you choose you will get more or less info.

As I stated before, it would be nice if someone could test this for a few minutes to see if something needs a fix.

Dezash:
> Because saving into a table won't query the info on the database unless the table doesn't have the info we want, sometimes all the info won't be displayed.

Why not? Isn't the table saving the same data as database (nickname, serial, ip and last connection)? Which info won't be displayed?

PlatinMTA:
> Why not? Isn't the table saving the same data as database (nickname, serial, ip and last connection)? Which info won't be displayed?

The way I made it is that you can store multiple names and multiple serials whenever they both don't have the same info, for example you could have:

1) Platin         AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA1
2) Platin         AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA2
3) Platin_        AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA1
4) Platin_        AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA2

The problem with this is that maybe you have 1, 3 and 4 stored on the table, but you don't have 2 so whenever you run "/whowas Platin" you will get only 1, because it's in the table. Meanwhile if you use "/whowas Platin_" you will get 3 and 4 with no problem.

If someone joined the server it will always find it, no matter what, but sometimes it won't display all the information that is known from the user.

This is the problem I face when trying to save the recent ones on a table and the rest on the database. However it should improve on the number of queries executed on a semi-busy server and I don't think it really bothers anyone to have slightly less info about an user that just recently joined that day.

Do you have any other idea that I could go through so we don't have to deal with this issue?

Dezash:
> I could make it search the user on the table, store it, and then make it search it on the database. Maybe that's better?

I think that would work. I also see an opportunity to improve updateWhowas by checking if any entries in aWhowas table match the new entry and update it instead of inserting a new one. It is better to do it on a table that contains ~500 entries than on the database that could contain millions.

PlatinMTA:
> I think that would work. I also see an opportunity to improve updateWhowas by checking if any entries in aWhowas table match the new entry and update it instead of inserting a new one. It is better to do it on a table that contains ~500 entries than on the database that could contain millions.

I will try to do that in the following days.

qaisjp:
> > I could make it search the user on the table, store it, and then make it search it on the database. Maybe that's better?
> 
> I think that would work. I also see an opportunity to improve updateWhowas by checking if any entries in aWhowas table match the new entry and update it instead of inserting a new one. It is better to do it on a table that contains ~500 entries than on the database that could contain millions.

didn't `whowas` give a history of instead of just the most recent field?

[admin]/admin2/server/admin_functions.lua
```diff
@@ -471,42 +471,112 @@ aFunctions = {
...
-		    outputConsole("Last serial #"..item..": "..row.serial, source)
-		    outputConsole("Last ip #"..item..": "..row.ip, source)
-		    outputConsole("Last connection #"..item..": "..last, source)
-	        end
-		return true
-	    else
-	        outputChatBox("We couldn't collect any info about \""..nick.."\".", source, 255, 141, 0)
-	        return false
-	    end
+            if not nick then outputChatBox("A nickname or serial is needed.", source, 255, 0, 0) return false end
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_functions.lua]: don't put on single line

[admin]/admin2/server/admin_functions.lua
```diff
@@ -471,42 +471,112 @@ aFunctions = {
...
+                        
+                        printF("Last ip #"..item..": "..data[3], source, 255, 141, 0)
+                        printF("Last connection #"..item..": "..last, source, 255, 141, 0)
+                    end
+                    return true
+                end
+            end
+            
+            if serial then
+                qh = dbQuery(db.connection, "SELECT * FROM whowas WHERE serial='"..nick.."';")
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_functions.lua]: sql injection here and elsewhere

[admin]/admin2/server/admin_server.lua
```diff
@@ -15,6 +15,46 @@ aInteriors = {}
...
+
+function clearRowsWhowas()
+    for k,data in ipairs(aWhowas) do
+        saveToDatabaseWhowas(data[1], data[2], data[3], data[4])
+    end
+    aWhowas = {}
+end
+
+function insertIntoWhowas(nick, serial, ip, timestamp)
+    if not nick or not serial or not ip then return end
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: don't put on single line

[admin]/admin2/server/admin_server.lua
```diff
@@ -15,6 +15,46 @@ aInteriors = {}
...
+function updateWhowas(player)
+    if not player or not isElement(player) then return end
+    
+    local nick = getPlayerName(player)
+    local serial = getPlayerSerial(player)
+    local ip = getPlayerIP(player)
+    local timestamp = getRealTime().timestamp
+    if get("whowassave") == "both" then
+        local rows = tonumber(get("whowasmax")) or 100
+        if rows > 500 then rows = 500 elseif rows < 5 then rows = 5 end
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: don't put on single line

[admin]/admin2/server/admin_server.lua
```diff
@@ -15,6 +15,46 @@ aInteriors = {}
...
+function insertIntoWhowas(nick, serial, ip, timestamp)
+    if not nick or not serial or not ip then return end
+    if not timestamp then timestamp = getRealTime().timestamp end
+    table.insert(aWhowas, {nick, serial, ip, timestamp})
+end
+
+function updateWhowas(player)
+    if not player or not isElement(player) then return end
+    
+    local nick = getPlayerName(player)
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: does `local nick, serial, ip = player.name, player.serial, player.ip` work?

[admin]/admin2/server/admin_server.lua
```diff
@@ -27,6 +67,11 @@ addEventHandler(
                 end
             end
             return
+        else
+            -- whowas stuff
+            for id, player in ipairs(getElementsByType("player")) do
+                updateWhowas(player)
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: might be better to just let `updateWhowas` take `root` for "all players"

then you can update `updateWhowas` to look like this:

```lua
function updateWhowas(player)
    if player == root then
        for _, p in ipairs(getElementsByType("player")) do
            updateWhowas(p)
        end
        return
    end
    
    ...
end
```

[admin]/admin2/server/admin_server.lua
```diff
@@ -76,6 +125,9 @@ addEventHandler(
             end
         end
         setPedGravity(source, getGravity())
+        
+        -- whowas stuff
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: this comment (and same ones elsewhere) are not helpful, the `updateWhowas` function is very clear in that does "whowas stuff"
```suggestion
```

[admin]/admin2/server/admin_server.lua
```diff
@@ -15,6 +15,46 @@ aInteriors = {}
...
+    aWhowas = {}
+end
+
+function insertIntoWhowas(nick, serial, ip, timestamp)
+    if not nick or not serial or not ip then return end
+    if not timestamp then timestamp = getRealTime().timestamp end
+    table.insert(aWhowas, {nick, serial, ip, timestamp})
+end
+
+function updateWhowas(player)
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]: `updateWhowas` is obvious that it deals with `whowas`, but it's not obvious _how_ and _what_ that update is.

Maybe `persistPlayerWhowasInfo` with a comment `-- Ensures that the player's whowas information is up to date.`

maybe something can also be said about ensuring db integrity here too

[admin]/admin2/server/admin_server.lua
Comment (PlatinMTA) [reply [admin]/admin2/server/admin_server.lua]: It does not.

[admin]/admin2/server/admin_server.lua
Comment (PlatinMTA) [reply [admin]/admin2/server/admin_server.lua]: Done

[admin]/admin2/server/admin_server.lua
Comment (PlatinMTA) [reply [admin]/admin2/server/admin_server.lua]: Deleted, sorry about that.

[admin]/admin2/server/admin_server.lua
Comment (PlatinMTA) [reply [admin]/admin2/server/admin_server.lua]: Done!

[admin]/admin2/server/admin_functions.lua
Comment (PlatinMTA) [reply [admin]/admin2/server/admin_functions.lua]: I think the change I made should fix this, if it does not please let me know.

PlatinMTA:
> didn't `whowas` give a history of instead of just the most recent field?

I can't really recall correctly, I think it gives a history. Whoever this whowas also does that, but it has a flaw that I described before.

I just updated the commits and made the changes that were requested. If more changes need to be done please let me know, but I would probably take some days to fix them since I'm working on something for my server.

[admin]/admin2/server/admin_server.lua
```diff
@@ -15,6 +15,66 @@ aInteriors = {}
...
+    if not nick or not serial or not ip then 
+        return 
+    end
+    if not timestamp then 
+        timestamp = getRealTime().timestamp 
+    end
+    table.insert(aWhowas, {nick, serial, ip, timestamp})
+end
+
+function persistPlayerWhowasInfo(player) -- Ensures that the player's information is up to date, it also ensures the integrity of the database
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_server.lua]:
```suggestion
-- Ensures that the player's information is up to date, it also ensures the integrity of the database
function persistPlayerWhowasInfo(player)
```

Comments should be above the function

[admin]/admin2/server/admin_functions.lua
```diff
@@ -469,6 +469,117 @@ aFunctions = {
...
+            local serial = false
+            local qh;
+            
+            if printF == "chat" then
+                printF = outputChatBox
+            else
+                printF = outputConsole
+            end
+            
+            if nick:len() == 32 then
```
Comment (qaisjp) [inline [admin]/admin2/server/admin_functions.lua]: Since serials are md5 hashes you can in fact do [`pregMatch(nick, '/^[a-f0-9]{32}$/') ~= false`](https://stackoverflow.com/questions/14300696/check-if-string-is-an-md5-hash)

Comment (PlatinMTA) [reply [admin]/admin2/server/admin_functions.lua]: I can't seem to make it work, any idea of what i'm doing wrong?

I use this:

`if nick:match("^[a-f0-9]{32}$") then`

Comment (sbx320) [reply [admin]/admin2/server/admin_functions.lua]: The regex @qaisjp posted is a perl style regex. `string.match` expects a Lua style regex. You'd need to use https://wiki.multitheftauto.com/wiki/PregMatch here.

However I'd advise against using a regex here, as regex is fairly slow, compared to just manually checking bytes.

Comment (PlatinMTA) [reply [admin]/admin2/server/admin_functions.lua]: Seeing as only an human will ever use the /whowas command I'll let the string lenght check.

Thanks @sbx320 for the help.

PlatinMTA:
I just updated the branch so it's compatible with clearchat, didn't test it tho.

Dutchman101:
@PlatinMTA why did you close the PR?

PR #164, opened 2019-08-07
admin: Fix isValidSerial

Dezash:
This PR fixes isValidSerial matching strings that are not necessarily valid.


[admin]/admin/server/admin_server.lua
```diff
@@ -286,9 +286,9 @@ function iif ( cond, arg1, arg2 )
 	return arg2
 end
 
+local serialExp = "^" .. string.rep ( "[A-F0-9]", 32 ) .. "$"
```
Comment (Citizen01) [inline [admin]/admin/server/admin_server.lua]: `[A-F0-9]` could have been `%x`
Source: https://www.lua.org/pil/20.2.html

[admin]/admin/server/admin_server.lua
```diff
@@ -1471,8 +1471,9 @@ addEventHandler ( "aBans", _root, function ( action, data, arg1, arg2, arg3 )
 			end
 		elseif ( action == "banserial" ) then
 			mdata = data
-			if ( isValidSerial ( data ) ) then
-				local newban = addBan ( nil,nil, string.upper ( data ),source,arg2, arg3 )
+			local serial = string.upper ( data )
+			if ( isValidSerial ( serial ) ) then
+				local newban = addBan ( nil,nil, serial,source,arg2, arg3 )
```
Comment (Citizen01) [inline [admin]/admin/server/admin_server.lua]: can you add spaces after all comas to seperate the args cleanly as well ? Thanks for the PR.

[admin]/admin/server/admin_server.lua
Comment (Dezash) [reply [admin]/admin/server/admin_server.lua]: @Citizen01 %x is equivalent to [0-9a-fA-F]. The idea was not to include lower case characters.

[admin]/admin/server/admin_server.lua
Comment (Dezash) [reply [admin]/admin/server/admin_server.lua]: Fixed.

[admin]/admin/server/admin_server.lua
Comment (Citizen01) [reply [admin]/admin/server/admin_server.lua]: seems fair.

qaisjp:
Thanks for the PR @Dezash and thanks for the review @Citizen01

PR #167, opened 2019-09-01
Changed renderKnifingTag() getElementPosition to getPedBonePosition

Lokote1998:
With this change the dxDrawText isn't just going to be static 1.3m above the player's position, this will follow the player's head.

Before: https://imgur.com/xSpGrdp
After: https://imgur.com/oCjwdsT


patrikjuvonen:
Thanks for the PR!

PR #170, opened 2019-09-09
showing player's country on join

Haxardous:


qaisjp:
1. does this mapping already exist in the admin resource? i think it should be moved there, with the full country name returned as the second result from `admin:getPlayerCountry`
2. `getCountry` should return false if the admin resource is not started
3. if false is returned from `getCountry`, you should just omit the country text. this will allow `joinquit` to work without a strict dependency on `admin`

Haxardous:
1) It doesn't, yeah it'll be functional lif it's on admin resource instead

PR #173, opened 2019-09-29
vehicles.maxidletime can be disabled now

Lord-Henry:
The setting vehicles.maxidletime now accepts negative values. Any value below 0 will disable this functionality and the vehicles created by Freeroam will no longer be destroyed by idle time.


DNL291:
I think it would be better disabling it by setting 0 instead of any negative value. Also, add a description for the setting in the meta.

Issue #111, opened 2018-07-25
admin2: decide content for "network" tab


qaisjp:
**Is your feature request related to a problem? Please describe.**
There is a "network" tab and some existing code ([client](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/admin2/client/main/admin_network.lua), [server](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/admin2/server/admin_network.lua)). **Let's brainstorm features to put into this tab.**

**Describe the solution you'd like**
This issue does not aim to _implement_ the Network tab. This issue is for us to discuss potential features we can include. A separate issue will be filed by an MTAQA member describing how the "Network" tab will function at a later date.

**What the tab looks like currently**
![image](https://user-images.githubusercontent.com/923242/43176160-7a616ac6-8fba-11e8-9bdd-57d49dda2721.png)

MIKI785:
I suppose it should contain info provided by [getNetworkStats](https://wiki.multitheftauto.com/wiki/GetNetworkStats) and maybe [getPerformanceStats](https://wiki.multitheftauto.com/wiki/GetPerformanceStats) (Packet usage category). And maybe an option to see this info per player.

alasmi8:
@MIKI785

I agree with you

jlillis:
Given the lack of progress here, I'm going to close this for now. The network tab was removed in #178.

PR #178, opened 2019-11-29
admin2: many ux improvements

### LINKED ISSUE #110: admin2: evaluate use of comboboxes
**Describe the bug**
The comboboxes used are actually buttons with some sort of dropdown image. This issue is to track the use of these custom comboboxes.

First task is to see whether we should move to a native combobox or fix the problems with the current system.

Second task would be to move to a new combobox, or fix the issues described below:

**Problems with the current system, and how to reproduce, and what is expected**

You can find a combobox on "Resources tab" -> "View type"

> _What is the problem?_
> You cannot click the button to open the combobox - you must click the dropdown image.
>
> _Why is this bad?_
> Poor UX
>
> _How to reproduce?_
> Click on the button, then click on the dropdown image

> _What is the problem?_
> Clicking the button brings it to the front. If you hold mouse down it will bring it to the front - letting go with mouse outside the button does not send it to the back (if you click the button without moving your mouse then it does send it to the back)
>
> _Why is this bad?_
> It is unexpected behaviour

> _What is the problem?_
> The gridlist is misplaced resulting in weird things being clicked sometimes
>
> _Why is this bad?_
> You would expect more accurate selection
>
> _How to reproduce?_
> "Double click slowly" on the dropdown image. Even though the "map" item is the second item down (for the Resources -> View type combobox) it will select that item.
>
> _What is expected?_
> Maybe make the button disabled, make the gridlist headerless, and push it down a little bit so it is not covering the top?
> Or you can remove the header and make it an empty item that is unselectable?

**Screenshots**
image
image

### LINKED ISSUE #113: admin2: navigate to players from search box
I find it frustrating having to click on the gridlist after searching from the search box!

Pressing down from the searchbox should go onto the gridlist. Pressing up on the first item should go onto the search box.

Press tab and shift+tab instead? Potential clash with changing tabs there. We should keep away from these shift+tab/tab keybinds until we have decided on an overall keybind plan.

**Additional context**
This is an **enhancement** to be worked on after feature parity is established (or when someone is super duper bored).

jlillis:
This pull request:

* Resolves #113 (up/down arrow can now navigate the player list while the search box has input focus)
* Resolves #110 by either:
     * replacing custom dropdowns with standard CEGUI comboboxes
     * replacing custom dropdowns with full widgets (see vehicle upgrade, give weapon, and give vehicle options)
* Adjusts positions of a few buttons on the player tab
* Adjusts the size of the admin messages view so the memo field does not overlap the close button
* Adds missing "Play sound on incoming admin chat message" option to the options tab and allows admin chat options to be modified from both the chat and options tabs
* Hides network tab for now (pending resolution on #111)
* Adds missing warp and warpto buttons

There are still a few things I'd like to address and this probably needs to be tested before it's really ready, so I'll leave it as a draft for now. In particular, I could use feedback on:
* The slap button - are multiple hp penalty options needed?
* The vehicle upgrades menu works, but I think the interface could use further improvement. The color options seem to use the old color palette system. Could the freeroam colorpicker be used here instead?
* There are some code quality improvements to be made with some of the widgets.


qaisjp:
This is some good work.

To answer your questions:

* > The slap button - are multiple hp penalty options needed?

  I don't know, to be honest. Parity with admin1 sounds good to me?

* > The vehicle upgrades menu works, but I think the interface could use further improvement. The color options seem to use the old color palette system. Could the freeroam colorpicker be used here instead?

  Good idea. I think for now this can be left alone and an issue can be created?

jlillis:
>       I don't know, to be honest. Parity with admin1 sounds good to me?
I'll see what I can do with slap options. The comboboxes have a minimum height larger than the height of the buttons, so I'll need to think up a way to implement it that isn't messy.

>       Good idea. I think for now this can be left alone and an issue can be created?
Sure, this PR is starting to get a little too large. I'll try to come up with a resolution for the slap options dropdown and make a code cleanup pass, then mark it as ready.

jlillis:
Moved some things around on the player tab and made the panel 20% large overall to make the slap options menu fit well:

image

I think this is ready for review. I'll cover code quality and other enhancements in separate PRs - this one is getting a little large.

PR #182, opened 2020-01-23
Optimised images

ricksterhd123:
*Total -- 687.97kb -> 586.16kb (14.8%)
Maybe not worth it?


Dutchman101:
Thanks for the PR, but i am already working on this, although the reduction will be greater than 14.8% and there will be no changed pixels at all. Even then, it will be limited since we also did it in the past (using an inferior method) at 99a39b7

So stay tuned!

ricksterhd123:
No problem, I am curious - is there any tools which can compress the images inside a .txd automatically?

dmi7ry:
> No problem, I am curious - is there any tools which can compress the images inside a .txd automatically?

It can't be done automatically right, because such tool can't know what format is requred by each texture (for an example, some of them should be unpacked ARGB, otherwise GTA will be crashed, etc).

Dutchman101:
Not that i know of, but for manually doing it, give this a read: [https://wiki.multitheftauto.com/wiki/Optimize_Custom_TXD](https://wiki.multitheftauto.com/wiki/Optimize_Custom_TXD)

PR #188, opened 2020-03-12
editor: Add OOP support for editor_test map

Lordmau5:
Heyo!

I've recently gotten back into MTA and am already a big fan of the OOP-syntax - sadly when creating maps in the editor said OOP-syntax isn't working...

After looking around the files a bit I noticed that the `editor_test` map it generates (more specifically the `meta.xml` for it) doesn't include the `<oop>true</oop>` node.

This PR fixes that.

Seeing as the old methods are still working even if this is on I don't see it as a downside at all - quite the contrary, it's only beneficial to those that want to use that new syntax! :)

### DIFF
--- a/[editor]/editor_main/server/dumpxml.lua
+++ b/[editor]/editor_main/server/dumpxml.lua
@@ -79,6 +79,11 @@ function dumpMeta ( xml, extraNodes, resource, filename, test )
 	if not resource then return false end
 	dimension = dimension or 0
 	extraNodes = extraNodes or {}
+
+	--Add OOP support
+	local oopNode = xmlCreateChild(xml, "oop")
+	xmlNodeSetValue(oopNode, "true")
+
 	--[[ info tag ]]--
 	local infoNode = xmlCreateChild(xml, "info")
 


Lordmau5:
I gotta do another PR for this as, apparently, I didn't incorporate the `cleanResource` method...

Every time the map is saved it adds *an extra* `oop` node 😁
Trying to figure out how to prevent that from happening, but yeah, thought I'd mention that as otherwise a meta.xml file could easily be filled up with tons of duplicate oop nodes...

qaisjp:
Thanks for the heads up. Please do that as soon as possible! Cheers

Lordmau5:
Tried doing it the same way that you guys are doing it for the `info` node but it doesn't want to properly remove it for whatever reason...

It only applies to the original resource - the `editor_test` resource seems to only have one OOP node.

Which method is in charge of saving to the original resource in case you know it off the top of your head?

qaisjp:
> Which method is in charge of saving to the original resource in case you know it off the top of your head?

Unfortunately I don't know!

PR #192, opened 2020-03-18
votemanager: added additional resource settings for ban votes

Dark-Dragon:
banPlayer provides a lot of banning options, none of which votemanager currently supports changing for its ban votes. This commit adds the following settings to the resource: 

1. voteban_banip true/false (default: true);
2. voteban_banusername true/false (default: false);
3. voteban_banserial true/false (default: false);
4. voteban_duration number in seconds (default: 3600)

Default values are the same as the default within banPlayer, except for duration, which is 1 hour instead of infinite. Note that voteban_enabled default remains false just as before.

The addition of the empty reason string when no reason is supplied via command is a workaround to avoid issues with lua's unpack() when unpacking tables containing nil entries. Here's an example: `local tab = {true,false,false,nil,5} tab.cheese = "Yes" local a,b,c,d,e = unpack (tab) outputChatBox(tostring(e))` should output 5 but instead outputs nil. Without the modification `tab.cheese = "Yes"` it will correctly output 5. Meanwhile directly referencing it without unpack always seems to be working as intended `local tab = {true,false,false,nil,5} tab.cheese = "Yes" local a,b,c,d,e = unpack (tab) outputChatBox(tostring(tab[5]))` outputting 5 correctly. Unless banPlayer internally handles an empty string very differently to nil I don't see any major drawbacks with this. I can't think of any better workaround, as votemanager relies on further modifying the tables and unpack() to read from them. But maybe someone else has a good idea.

Lastly note that votekicks weren't functional before and will not be before https://github.com/multitheftauto/mtasa-blue/issues/1301 gets resolved.

### DIFF
--- a/[managers]/votemanager/meta.xml
+++ b/[managers]/votemanager/meta.xml
@@ -48,9 +48,9 @@
 		<setting name="*voteban_locktime" value="[120]"/>
 		<setting name="*voteban_percentage" value="[75]"/>
 		<setting name="*voteban_allowchange" value="[true]"/>
-		<setting name="*voteban_banip" value="[true]"/>
+		<setting name="*voteban_banip" value="[false]"/>
 		<setting name="*voteban_banusername" value="[false]"/>
-		<setting name="*voteban_banserial" value="[false]"/>
+		<setting name="*voteban_banserial" value="[true]"/>
 		<setting name="*voteban_duration" value="[3600]"/>
 
 		<setting name="*votekill_enabled" value="[false]"/>
--- a/[managers]/votemanager/meta.xml
+++ b/[managers]/votemanager/meta.xml
@@ -48,6 +48,10 @@
 		<setting name="*voteban_locktime" value="[120]"/>
 		<setting name="*voteban_percentage" value="[75]"/>
 		<setting name="*voteban_allowchange" value="[true]"/>
+		<setting name="*voteban_banip" value="[true]"/>
+		<setting name="*voteban_banusername" value="[false]"/>
+		<setting name="*voteban_banserial" value="[false]"/>
+		<setting name="*voteban_duration" value="[3600]"/>
 
 		<setting name="*votekill_enabled" value="[false]"/>
 		<setting name="*votekill_timeout" value="[15]"/>
--- a/[managers]/votemanager/votemanager_polls.lua
+++ b/[managers]/votemanager/votemanager_polls.lua
@@ -58,6 +58,12 @@ addEventHandler("onResourceStart", thisResourceRoot,
 			info.percentage  = get(settingsGroup.."_percentage") or nil
 			info.timeout     = get(settingsGroup.."_timeout")    or nil
 			info.allowchange = get(settingsGroup.."_allowchange")
+			if name == "ban" then
+				info.banip       = get(settingsGroup.."_banip")
+				info.banusername = get(settingsGroup.."_banusername")
+				info.banserial   = get(settingsGroup.."_banserial")
+				info.duration    = get(settingsGroup.."_duration")
+			end
 			info.blockedPlayers = {}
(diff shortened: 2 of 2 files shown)


[managers]/votemanager/meta.xml
```diff
@@ -48,6 +48,10 @@
 		<setting name="*voteban_locktime" value="[120]"/>
 		<setting name="*voteban_percentage" value="[75]"/>
 		<setting name="*voteban_allowchange" value="[true]"/>
+		<setting name="*voteban_banip" value="[true]"/>
+		<setting name="*voteban_banusername" value="[false]"/>
+		<setting name="*voteban_banserial" value="[false]"/>
+		<setting name="*voteban_duration" value="[3600]"/>
```
Comment (Dark-Dragon) [inline [managers]/votemanager/meta.xml]: adding the 4 settings, the chosen setting names should be understandable intuitively

[managers]/votemanager/votemanager_polls.lua
```diff
@@ -58,6 +58,12 @@ addEventHandler("onResourceStart", thisResourceRoot,
 			info.percentage  = get(settingsGroup.."_percentage") or nil
 			info.timeout     = get(settingsGroup.."_timeout")    or nil
 			info.allowchange = get(settingsGroup.."_allowchange")
+			if name == "ban" then
```
Comment (Dark-Dragon) [inline [managers]/votemanager/votemanager_polls.lua]: not entirely sure whether this or `if settingsGroup == "voteban" then` would be better. I think either is fine.

[managers]/votemanager/votemanager_polls.lua
```diff
@@ -426,6 +432,8 @@ function voteKick(player, reason)
 		local title = "Kick "..getPlayerName(player).."?"
 		if reason then
 			title = title.." ("..reason..")"
+		else
+			reason = ""
```
Comment (Dark-Dragon) [inline [managers]/votemanager/votemanager_polls.lua]: making reason an empty string if none is supplied for reasons explained in the pull request

[managers]/votemanager/votemanager_polls.lua
```diff
@@ -446,14 +454,16 @@ function voteBan(player, reason)
 		local title = "Ban "..getPlayerName(player).."?"
 		if reason then
 			title = title.." ("..reason..")"
+		else
+			reason = ""
```
Comment (Dark-Dragon) [inline [managers]/votemanager/votemanager_polls.lua]: making reason an empty string if none is supplied for reasons explained in the pull request

[managers]/votemanager/votemanager_polls.lua
```diff
@@ -446,14 +454,16 @@ function voteBan(player, reason)
...
 		end
 		return startPoll{
 			title=title,
 			percentage = vote.ban.percentage,
 			visibleTo = rootElement,
 			timeout = vote.ban.timeout,
 			allowchange = vote.ban.allowchange;
-			[1]={"Yes",banPlayer,player,serverConsole,reason},
+			[1]={"Yes",banPlayer,player,vote.ban.banip,vote.ban.banusername,vote.ban.banserial,serverConsole,reason,vote.ban.duration},
```
Comment (Dark-Dragon) [inline [managers]/votemanager/votemanager_polls.lua]: if reason was nil here it would result in the ban duration being ignored, because after unpacking the table in https://github.com/multitheftauto/mtasa-resources/blob/38d70861bc62c826772e82df879196851d4a5826/%5Bmanagers%5D/votemanager/votemanager_server.lua#L313 it would get lost

jlillis:
This PR appears to work as intended, but I think it would be better for players to be banned by serial by default. Thoughts?

xLive:
and I think IP ban should be disabled by default? 

I know some people play with each other and have the same IP
I don't think it's fair to ban all of them? and serial is enough

Dark-Dragon:
I chose the default to be consistent with the default of [banPlayer](https://wiki.multitheftauto.com/wiki/BanPlayer) as well as how voteban previously banned players, but if everyone agrees that serial is the way to go I see no problem with changing it to reflect that.

PR #194, opened 2020-03-25
freeroam: fix exploit to get around hex code removal

zo3n:
For a long time it has been possible to avoid hex removal from a message in Freeroam. This PR fixes the problem.

The following message:
#00ff00color #00ff#ffffff00code

"color" would have it's hex code removed, and so would "code", but after removal of one hex code on word "code", it would join together as #00ff00code and thus the hex removal was avoided.

My apologies on description though. This PR makes sure that there can be no hex code present in player's message and fixes the earlier abuse problem.

### DIFF
--- a/[gameplay]/freeroam/fr_server.lua
+++ b/[gameplay]/freeroam/fr_server.lua
@@ -443,7 +443,10 @@ addEventHandler('onPlayerChat', root,
 			end
 			if isElement(source) then
 				local r, g, b = getPlayerNametagColor(source)
-				outputChatBox(getPlayerName(source) .. ': #FFFFFF' .. msg:gsub('#%x%x%x%x%x%x', ''), root, r, g, b, true)
+				while(msg:gsub('#%x%x%x%x%x%x', ''):len() ~= msg:len()) do
+					msg = msg:gsub('#%x%x%x%x%x%x', '')
+				end
+				outputChatBox(getPlayerName(source) .. ': #FFFFFF' .. msg, root, r, g, b, true)
 				outputServerLog( "CHAT: " .. getPlayerName(source) .. ": " .. msg )
 			end
 		end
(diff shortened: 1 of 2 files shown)


[gameplay]/freeroam/fr_server.lua
```diff
@@ -443,7 +443,10 @@ addEventHandler('onPlayerChat', root,
 			end
 			if isElement(source) then
 				local r, g, b = getPlayerNametagColor(source)
-				outputChatBox(getPlayerName(source) .. ': #FFFFFF' .. msg:gsub('#%x%x%x%x%x%x', ''), root, r, g, b, true)
+				while(msg:gsub('#%x%x%x%x%x%x', ''):len() ~= msg:len()) do
+					msg = msg:gsub('#%x%x%x%x%x%x', '')
+				end
```
Comment (qaisjp) [inline [gameplay]/freeroam/fr_server.lua]: You could do something like

```lua
function stripHex(str)
    local oldLen
    repeat
        oldLen = str:len()
        str = str:gsub('#%x%x%x%x%x%x', '')
    until str:len() == oldLen
    return str
end
```

and then call that function

Comment (zo3n) [reply [gameplay]/freeroam/fr_server.lua]: Alright, makes sense. Should the function be in **fr_server.lua** or **util.lua**?
Also, with a while loop your function can be shorter:

```lua
function stripHex(str)
    while(str:gsub('#%x%x%x%x%x%x', ''):len() ~= str:len()) do
        str = str:gsub('#%x%x%x%x%x%x', '')
    end
    return str
end
```

Unless you're aiming for more understanding of the function? Which function should be used?

Comment (qaisjp) [reply [gameplay]/freeroam/fr_server.lua]: I'm aiming for more understanding + preventing `gsub` from being called more than necessary.

Comment (zo3n) [reply [gameplay]/freeroam/fr_server.lua]: Should be okay now

PR #213, opened 2020-06-15
Fix min_mta_version warning in missiontimer

jlillis:
This PR fixes the following warning in missiontimer (used by a few of the default gamemodes):

`WARNING: [mtasa-resources]\[gameplay]\missiontimer\missiontimer_server.lua:29: <min_mta_version> section in the meta.xml is incorrect or missing (expected at least server 1.3.0-9.04570 because a send list is being used)`


qaisjp:
danke

PR #219, opened 2020-06-19
Fix for #175

### LINKED ISSUE #175: EDF Can't open properties window for object element
editor

**Describe the bug**
 Well you can't open the properties window by double click or other way

**To Reproduce**
just create s simple edf file like this
```xml
<def name="test">
    <element name="custom_object" friendlyname="Custom Object" icon="edf/object.png">
        <data name="model" type="objectID" default="1271" />
        <data name="position" type="coord3d" default="0,0,0" />
        <data name="rotation" type="coord3d" default="0,0,0" />
        <object model="!model!" position="!position!" rotation="!rotation!" />
    </element>
</def>
```
then spawn custom_object and try to double click nothing happens only with object

arfni:
Fixes #175 , the only solution that I could come up with for this issue that most likely won't break anything. (As I don't know in which cases it might be necessary to remove collisions for an elements representations)

It now iterates through the elements children and only disables the collisions if it has a representation that is not an object.

### DIFF
--- a/[editor]/editor_main/client/main.lua
+++ b/[editor]/editor_main/client/main.lua
@@ -637,8 +637,8 @@ end
 
 function requiresObjectCollisions(element)
 	for k, child in ipairs(getElementChildren(element)) do
-		if getElementData(child, "edf:dummy") == false then
-			if getElementData(child, "edf:rep") == true then
+		if not getElementData(child, "edf:dummy") then
+			if getElementData(child, "edf:rep") then
 				if getElementType(child) ~= "object" then
 					return false
 				end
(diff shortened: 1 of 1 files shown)


[editor]/editor_main/client/main.lua
```diff
@@ -635,6 +635,19 @@ function setRepresentationCollisionsEnabled(element, state)
 	end
 end
 
+function requiresObjectCollisions(element)
+	for k, child in ipairs(getElementChildren(element)) do
+		if getElementData(child, "edf:dummy") == false then
+			if getElementData(child, "edf:rep") == true then
```
Comment (qaisjp) [inline [editor]/editor_main/client/main.lua:641]: I'd use "implicit" value checks here (`if x then`, `if not x then`) instead but other than that, this LGTM.

qaisjp:
Please can you confirm this is tested btw, and if so, I can merge this

arfni:
I did test it.

But as i mentioned:
> (As I don't know in which cases it might be necessary to remove collisions for an elements representations)
Though I think if there are any cases, they'll work fine.

PR #220, opened 2020-06-20
editor: replace 1729ceb (#219) with a better fix

arfni:
When dragging the representation the cursor could point at it which would result in the object moving away from itself.
For whatever reason I didn't come across this when  testing.
_And now I know why the collisions are  being disabled._

So here's a better fix that only disables the collisions when the element is being dragged.


qaisjp:
Thanks!

PR #221, opened 2020-06-23
Add admin/conf files to .gitignore

galeapatrik:
.gitignore
+ ~~[admin]/admin/conf/IpToCountryCompact.csv~~
+ [admin]/admin/conf/reports.xml

File deletions
- [admin]/admin/conf/reports.xml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,10 @@
 
 \[test\]
 
+# admin
+\[admin\]/admin/conf/IpToCountryCompact.csv
+\[admin\]/admin/conf/reports.xml
+
 # admin2
 \[admin\]/admin2/admin.db
 \[admin\]/admin2/conf/reports.xml
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 \[test\]
 
 # admin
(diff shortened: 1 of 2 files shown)


qaisjp:
The files needs to be deleted in the repo as well. (Adding it to the `.gitignore` does not automatically delete the file from the repository.)

Also, the `IpToCountryCompact.csv` file was deliberately added in 9e2005f05b007ceabbb63d2448f7735dd47dae57 - so we need to think of a way to ensure that this `csv` file is added to the official `admin.zip`s.

I've outlined my solution to this in #222.

galeapatrik:
> The files needs to be deleted in the repo as well. (Adding it to the `.gitignore` does not automatically delete the file from the repository.)
> 
> Also, the `IpToCountryCompact.csv` file was deliberately added in [9e2005f](https://github.com/multitheftauto/mtasa-resources/commit/9e2005f05b007ceabbb63d2448f7735dd47dae57) - so we need to think of a way to ensure that this `csv` file is added to the official `admin.zip`s.
> 
> I've outlined my solution to this in #222.

Okay, maybe later I try to move **ip2c** to a different resource.
Until that I revoke the IpToCountryCompact.csv ignore.

PR #223, opened 2020-06-24
deathmatch: major refactor

jlillis:
When merged, this PR will refactor the `deathmatch` gamemode. Here is a list of improvements:

- Integration with `defaultstats` to handle default weapon stats
- Integration with `mapcycler` to trigger a map vote at end-of-round
- The gamemode will now respect the `deathpickups.only_current` setting
- More robust handling of game/player states
- New loading screen that displays the map name and author
- Game logic regarding the HUD and camera moved clientside
- General code quality improvements, including comments
- Line endings converted to LF - this refactor turned more into a rewrite, so I chose to treat the code as new. As discussed on discord new files are supposed to use LF endings. Unfortunately, this makes some of the diffs a little messy.
- Redesigned HUD designed to emulate vanilla style
- Added ability to define infinite frag or time limits (not both simultaneously) and disable respawn

This refactored version is fully compatible with existing deathmatch maps. I believe it is at a point where it is ready for feedback and testing.

Some TODOs:
- Improvements to textlib code for performance (use rendertargets). Improvements to textlib code should be ported to other resources using textlib.
- Improvements to existing deathmatch maps - maybe add some new ones?


Dutchman101:
@jlillis This is still a draft right? Any plans to continue, or are you waiting for others to do so? (considering it was almost a year ago and this PR is stale)

jlillis:
There are still a few changes I've made but not pushed to a PR yet. I'll have this updated and moved out of draft in a week or two.

jlillis:
I have a few other PRs that I would ideally like to get merged ahead of this, but I'm going to say that this is ready for review at least.

[gamemodes]/[deathmatch]/deathmatch/client/main.lua
```diff
@@ -0,0 +1,155 @@
...
+    exports.scoreboard:scoreboardSetColumnPriority("Rank", 1)
+    exports.scoreboard:scoreboardSetSortBy("Rank")
+    -- fade out camera
+    fadeCamera(false, 0)
+    -- if a game is in progress, apply the loading camera matrix
+    if getElementData(resourceRoot, "gameState") == GAME_IN_PROGRESS then
+        setCameraMatrix(unpack(calculateLoadingCameraMatrix()))
+    end
+    -- inform server we are ready to play
+    triggerServerEvent("onDeathmatchPlayerReady", localPlayer)
```
Comment (xLive) [inline [gamemodes]/[deathmatch]/deathmatch/client/main.lua:17]: Change the source of this event to `resourceRoot`? 
> Note: To save server CPU, you should avoid setting theElement to the root element where possible. Using resourceRoot is usually sufficient if the event is handled by the same resource on the server.

Comment (srslyyyy) [reply [gamemodes]/[deathmatch]/deathmatch/client/main.lua:17]: localPlayer is lower in element tree. You can go for resourceRoot (not everytime), but ideally you should be using localPlayer instead.

Comment (Pirulax) [reply [gamemodes]/[deathmatch]/deathmatch/client/main.lua:17]: Yes, when triggering the event you want to trigger it at the lowest possible point in the element tree.

jlillis:
Got this cleaned up. I think it's ready to merge but hasn't been playtested beyond me and a VM. If this looks good, I'd like to merge and then give tdm/tdma modes the same treatment.

Dutchman101 (approved):

Issue #222, opened 2020-06-23
New `ip2c` resource


qaisjp:
**Is your feature request related to a problem? Please describe.**

The worktree gets dirty whenever we start the admin resource, because the admin resource occasionally generates a new `IpToCountryCompact.csv`. We can't delete that file (#221) because sometimes servers can't actually generate the file themselves (9e2005f05b007ceabbb63d2448f7735dd47dae57).

Other resources (the only non-default one is `dxscoreboard`) is unnecessarily dependent on `admin` to provide the country flags feature.

**Describe the solution you'd like**

**TL;DR:** We create a new `ip2c` resource without this data file in our repository. We generate the file when creating the main resources archive. `dxscoreboard`, `admin`, and `admin2` will now auto-start `ip2c` if it exists, and function without country detection otherwise.

We can generate the `IpToCountryCompact.csv` file **on the build server** when creating our release archive—either by reimplementing the code in Python, or building some sort of Lua wrapper that implements some of MTA's native functions using publicly available [rocks](https://luarocks.org/) [0].

I'd prefer the latter solution instead of the former as then we don't need to duplicate the code or put effort into keeping it consistent. The question is how flexible the admin code currently is—hopefully this is not an issue as we just move `ip2c` to its own resource.

Instead of `dxscoreboard` and `admin` hard-failing when `<include>` can't find `ip2c`, we do some checks and manually start `ip2c` if it exists. Or we add an `optional` field to `include` instead of implementing the same logic everywhere. (Use case: what if the server author would like to keep `ip2c` _loaded_, but not have it auto-start when e.g. `admin` starts?)

**Additional context**

It would also be nice to use this as an excuse to GitHubify our build process here. It would be nice to move the mtasa-resources build task off the build server, and into GitHub Actions. /cc @ccw808

[0] Note that LuaRocks support will not be necessary inside MTA. It would only be used on the build server to generate the csv file.

qaisjp:
Oh, looking at the code (why didn't I do this earlier?), `admin` just fetches the csv file from MTA's servers. And I assume MTA's server just generates the csv using some script. It'd take just a couple seconds to add `wget http://mirror.multitheftauto.com/mtasa/scripts/IpToCountryCompact.csv` to the build steps.

See:

https://github.com/multitheftauto/mtasa-resources/blob/e51ed2fb1736c547ca0c7984d43a161d2c6872ac/%5Badmin%5D/admin/server/admin_ip2c.lua#L12-L14

[...]

PR #224, opened 2020-06-24
Separate ip2c resource

Patrick2562:
issue #222


galeapatrik:
just revert e3b5a4e to use old one (admin2's ip2c)

jlillis:
I think we should consider closing this for now until a consensus is reached in #222

PR #225, opened 2020-06-24
playerblips: various improvements

jlillis:
When merged, this PR will make various improvements to the `playerblips` resource:

- Refactored settings config so they appear in `admin` and `admin2`. This *breaks backwards compatibility* as the setting names had to be changed. The major version of the resource was bumped accordingly.
- Added `*blip_range` setting to adjust the range at which blips are shown on the minimap
- Added optional `playercolors` integration and a new setting: `*use_nametag_colors`
- Modified `freeroam` to use this resource for player blip functionality, rather than it's own code.
- Fixed a debug error that occurs when other resources (i.e race) destroy all blips attached to players
- CRLF -> LF: this makes the final diff look weird: just follow along from 5853a42


jlillis:
I'm happy with this and would like to merge, but would appreciate a second set of eyes per repo policy.

[gameplay]/playerblips/server/main.lua
```diff
@@ -0,0 +1,71 @@
+local useTeams = get("*use_team_colors") == "true" and true or false
+local useNametags = get("*use_nametag_colors") == "true" and true or false
```
Comment (qaisjp) [inline [gameplay]/playerblips/server/main.lua:2]:
```suggestion
local useTeams = get("*use_team_colors") == "true"
local useNametags = get("*use_nametag_colors") == "true"
```
Nit: this is the same

[gameplay]/playerblips/server/main.lua
```diff
@@ -0,0 +1,71 @@
...
+	if isElement(blips[player]) then
+		blips[player]:destroy()
+	end
+	blips[player] = nil
+	colors[player] = nil
+end
+
+addEventHandler("onPlayerQuit", root, createPlayerBlip)
+addEventHandler("onPlayerWasted", root, createPlayerBlip)
+addEventHandler("onPlayerSpawn", root, createPlayerBlip)
```
Comment (qaisjp) [inline [gameplay]/playerblips/server/main.lua:71]: These events set the player to `source`, but `createPlayerBlip` does not use `source`.

qaisjp (commented):
Generally looks good but I think I found a problem

[gameplay]/playerblips/meta.xml
```diff
@@ -1,11 +1,55 @@
...
-	<settings>
-		<setting name="use_team_colors" value="false"/> 		<!-- Set to true to enable, anything else will not use teams. Default: false -->
-		<setting name="blip_color" value="[ [ 0, 255, 0 ] ]"/> 	<!-- Default blip color if se_team_colors is not enabled. Default: [ [ 0, 255, 0 ] ] -->
-		<setting name="blip_size" value="2"/> 					<!-- Size of a player blip. Default: 2 -->
-		<setting name="blip_alpha" value="255"/> 				<!-- Alpha value of a player blip (0 - 255). Default: 255 -->
-	</settings>
-	<oop>true</oop>
-</meta>
+<meta>
+	<info name="Player Blips" author="Noki" version="2.0.0" type="script" description="Attaches blips to players."/>
```
Comment (qaisjp) [inline [gameplay]/playerblips/meta.xml:2]: Also, feel free to add yourself here

[gameplay]/playerblips/server/main.lua
Comment (jlillis) [reply [gameplay]/playerblips/server/main.lua]: Fixed

[gameplay]/playerblips/server/main.lua
Comment (jlillis) [reply [gameplay]/playerblips/server/main.lua]: Not sure how I let this slide, but these event handlers were set up all wrong. Fixed.

[gameplay]/playerblips/meta.xml
Comment (jlillis) [reply [gameplay]/playerblips/meta.xml]: Fixed.

qaisjp (approved):

Issue #227, opened 2020-06-26
race: add car fade addon


AfuSensi:
**Is your feature request related to a problem? Please describe.**
Car fade is used by a lot of race servers, so it makes sense including it in race's default add-ons.

**Describe the solution you'd like**
Add car fade to the race add-ons 

**Additional context**
For those who don't know, car fade lowers other vehicle alpha based on proximity.
A lot of race servers use this to have better visibility while in ghost mode.

PR #228, opened 2020-06-26
race: add car fade addon

AfuSensi:
Resolves https://github.com/multitheftauto/mtasa-resources/issues/227
Since many race servers use car fade, adding it to the default race addons makes sense.
So here's my implementation.

Settings included:
- Maximum car fade distance
- Minimum car fade distance
- Minimum alpha
- Car fade keybind
- Can car fade be toggled
- Enabled by default
- Broadcast message at map start

carfade


[gamemodes]/[race]/[addons]/race_carfade/util_client.lua
```diff
@@ -0,0 +1,17 @@
...
+		for i,k in ipairs(keys) do
+			result[k] = vals
+		end
+	end
+	return result
+end
+
+function mathClamp(number, min, max)
+	return number <= min and min or (number >= max and max or number)
+end
```
Comment (qaisjp) [inline [gamemodes]/[race]/[addons]/race_carfade/util_client.lua:17]: Maybe just make this file `util_shared.lua` and take these functions out of `util_server.lua`?

botder:
I suggest you add the option to limit the carfade update rate. In the current state, the update runs on every onClientPreRender callback. My suggestions is to limit the update rate to 16 times per second max, like a frame limiter. Visually it should remain pretty much the same and you reduce CPU load.

AfuSensi:
I have updated a few things.
Added throttle setting. https://github.com/multitheftauto/mtasa-resources/pull/228#issuecomment-650124509
Changed util files. https://github.com/multitheftauto/mtasa-resources/pull/228#discussion_r446061063
Handling of vehicle `race.alpha` and player `race.alpha` independently since they can be set independently by race.

iiN3xT:
Good job! would be nice if you could extend it more and implement car hide as well

jlillis:
This looks like a great addition and I'd like to get this merged soon, but I'm not able to test it myself currently. Could someone try this on a server with some other players and report back with their results? If everything works I'll just send it.

jlillis (approved):
Code looks good and I've tested it - but on a server with just me on it. Going to yeet this since it's been sitting for so long.

PR #229, opened 2020-07-16
freeroam: Fix warning when time is set to freeze

LosFaul:
Fix warning when time is set to freeze in fr_client.lua

grafik

### DIFF
--- a/[gameplay]/freeroam/fr_client.lua
+++ b/[gameplay]/freeroam/fr_client.lua
@@ -1871,14 +1871,16 @@ addCommandHandler('st', setTimeCommand)
 function toggleFreezeTime()
 	local state = guiCheckBoxGetSelected(getControl(wndMain, 'freezetime'))
 	guiCheckBoxSetSelected(getControl(wndMain, 'freezetime'), not state)
-	setTimeFrozen(state)
+	local h, m = getTime()
+	local w = getWeather()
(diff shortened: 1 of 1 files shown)


[gameplay]/freeroam/fr_client.lua
```diff
@@ -1871,14 +1877,16 @@ addCommandHandler('st', setTimeCommand)
 function toggleFreezeTime()
 	local state = guiCheckBoxGetSelected(getControl(wndMain, 'freezetime'))
 	guiCheckBoxSetSelected(getControl(wndMain, 'freezetime'), not state)
+	freezeTimeHour, freezeTimeMinute = getTime()
+	freezeTimeWeather = getWeather()
 	setTimeFrozen(state)
```
Comment (qaisjp) [inline [gameplay]/freeroam/fr_client.lua:1882]: Isn't the fix just to do `setTimeFrozen(state, freezeTimeHour, freezeTimeMinute, freezeTimeWeather)` and there's no need for a global? Or does this also support changing time/weather while frozen (and it will freeze to the new time/weather).

Comment (LosFaul) [reply [gameplay]/freeroam/fr_client.lua:1882]: someone was complaining that it would change time/weather back few seconds later to the state freeze was enbaled
so, i added the 3 variables to keep the state updated when time and weather get changed while freeze is enabled

PR #234, opened 2020-08-19
editor: Fix creation of colshapes as children in custom EDFs

Yamsha75:
I've noticed that collision shapes are not created properly by editor resource and found the reason.

According to [mtawiki/Resource:Editor/EDF](https://wiki.multitheftauto.com/wiki/Resource:Editor/EDF) and file [editor/edf/datatypes.lua](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/edf/datatypes.lua#L169) valid types of collision shapes are:
```
One of: "colcircle", "colcube", "colrectangle", "colsphere", "coltube"
```

But in [[editor]/edf/edf.lua](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/edf/edf.lua#L93) there are checks for different strings - without the "col" in the beginning.

Also I've noticed that instead of ```createColCuboid``` a deprecated function ```createColCube``` is used.

Proposed PR fixes these and colshapes are created.

### DIFF
--- a/[editor]/edf/edf.lua
+++ b/[editor]/edf/edf.lua
@@ -97,7 +97,7 @@ local edfCreateBasic = {
 			return createColTube(cdata.position[1], cdata.position[2], cdata.position[3], cdata.radius, cdata.height)
 		elseif cdata.type == "rectangle" or cdata.type == "colrectangle" then
 			return createColRectangle(cdata.position[1], cdata.position[2], cdata.position[3], cdata.width, cdata.depth)
-		elseif cdata.type == "cube" or cdata.type == "cuboid" or cdata.type == "colcube" then
+		elseif cdata.type == "cube" or cdata.type == "cuboid" or cdata.type == "colcube" or cdata.type == "colcuboid" then
 			return createColCuboid(cdata.position[1], cdata.position[2], cdata.position[3], cdata.width, cdata.depth, cdata.height)
 		end
 	end,
--- a/[editor]/edf/edf.lua
+++ b/[editor]/edf/edf.lua
@@ -91,14 +91,14 @@ local edfCreateBasic = {
 		return createRadarArea(cdata.posX, cdata.posY, cdata.sizeX, cdata.sizeY, cdata.colorR, cdata.colorG, cdata.colorB, cdata.colorA)
 	end,
 	colshape = function(cdata)
-		if cdata.type == "sphere" then
+		if cdata.type == "sphere" or cdata.type == "colsphere" then
 			return createColSphere(cdata.position[1], cdata.position[2], cdata.position[3], cdata.radius)
-		elseif cdata.type == "tube" then
+		elseif cdata.type == "tube" or cdata.type == "coltube" then
 			return createColTube(cdata.position[1], cdata.position[2], cdata.position[3], cdata.radius, cdata.height)
-		elseif cdata.type == "rectangle" then
+		elseif cdata.type == "rectangle" or cdata.type == "colrectangle" then
 			return createColRectangle(cdata.position[1], cdata.position[2], cdata.position[3], cdata.width, cdata.depth)
-		elseif cdata.type == "cube" or cdata.type == "cuboid" then
-			return createColCube(cdata.position[1], cdata.position[2], cdata.position[3], cdata.width, cdata.depth, cdata.height)
+		elseif cdata.type == "cube" or cdata.type == "cuboid" or cdata.type == "colcube" then
+			return createColCuboid(cdata.position[1], cdata.position[2], cdata.position[3], cdata.width, cdata.depth, cdata.height)
 		end
 	end,
 	ped = function(cdata)


qaisjp (approved):
thank you! this looks good. have you tested this?

i'm happy to merge this if someone else can confirm this works :)

Yamsha75:
Haven't tested with all types yet. I will _soon_ provide a map file with colshapes and screenshots before and after these changes so there are no uncertainties

Yamsha75:
Ok there's more to this topic than I originally thought. My problem was that I wanted to create a new element type with colshapes as children in an edf for a custom gamemode and I went with something like this:
```xml
<element name="trailer_spawn" friendlyname="trailer spawnpoint" >
    <data name="depot" type="element:depot" />
    <data name="position" type="coord3d" default="0,0,0" />
    <data name="rotation" type="coord3d" default="0,0,0" />
    <vehicle editorOnly="true" model="450" alpha="100" />
    <colshape type="colsphere" posY="5" radius="1.5"/>
    <colshape type="colsphere" posY="-2" radius="1.5"/>
</element>
```

This works after proposed fix, seen here:
image

But after trying to add new colshapes by editing a map file I found that the function responsible for creating those is not fired. After looking around some more I found that element definitions from [[editor]/edf/edf.edf](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/edf/edf.edf#L67) are not the same as in [[editor]/editor_main/editor_main.def](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/editor_main/editor_main.edf), so I went to investigate further.


I have tried simply copying colshape element definition from edf resource to editor_main and that resulted in a new element type in editor gui, although creating a new colshape this way works a little different than creating a marker, for example. With a marker right after clicking the creation button you get a default marker attached to the cursor but with a colshape it's different: element properties gui as seen below shows up. After accepting the same gui window reappears with properties of the newly created colshape. The colshape is listed in editor's elements browser.
image
image

Saving resulted in this within the map file:
```xml
<colshape id="colshape (1)" width="10" height="10" depth="10" type="colcube" dimension="0" alpha="255" interior="0" posX="0" posY="0" posZ="10" rotX="0" rotY="0" rotZ="0"></colshape>
```

Some errors arose because element was not properly created in line 603 of [[editor]/edf/edf.lua](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/edf/edf.lua#L603) and ```theElement``` variable returned nil. No idea why, haven't investigated enough.

So far it looks somewhat good, but after loading the map, the shape cannot be seen anymore. I can find it in runcode by ```getElementsByType("colshape")[1]``` and check its position, but [getElementsWithinColShape](https://wiki.multitheftauto.com/wiki/GetColShapeSize) claims the element is not a colshape and method :getSize() does not exist. A new colshape created in either editor or runcode works just fine with those. Something is not OK with recreating colshapes from a map file.

Out of curiosity I also added other element definitions that were in edf.edf but not in editor_main.edf:
- blip - works fine with gui and everything
- radararea - is bugged: the new radar area is stuck to the cursor no matter what
- water - doesn't work at all

All were missing icons, as those aren't in [[editor]\editor_main\client\images\icons](https://github.com/multitheftauto/mtasa-resources/tree/master/%5Beditor%5D/editor_main/client/images/icons)
All in all to me it seems like implementing those in the editor resource was omitted or not finished.

jlillis:
This PR fixes the issue as described and I could merge it now, but are you able to/interested in adding the additional editor GUI options as shown in your screenshots? I think it would be a good change.

Yamsha75:
I guess this fixes the issue I described, but adding colshapes, radararea, blip and water as elements placeable in base editor would be an obvious benefit. Maybe a separate issue would be a good idea?

As to adding these features, I'm not sure I could do this on my own without any documentation and/or help from someone who knows more about editor/edf.

jlillis:
I would say just adding colshapes as placeable elements in the base editor would be in-scope for this PR. radarareas, blips, and water should be done in seperate PRs.

This screenshot you provided seems to suggest you already have colshapes as placeable elements - am I mistaken?
image

Yamsha75:
Well they are placeable, but there are issues:
- detaching the colshape from cursor is not working, you have to create a new element for it to detach, but sometimes it does not work and map reload is required
- later picking it up back again is impossible - no in-game collision is to blame I think, maybe modifying the event that is fired when client clicks left mouse button in editor and some line of sight processing will do
- element properties gui is a little broken because at colshape creation it is shown three times after each click of the accept button
- changing a colshape's type after placing it is not working

However, using colshapes as children in custom element definitions is working properly. Placing them as independent elements is bugged.

So yeah, kinda works, but I wouldn't say its ready to be merged. 😄

jlillis:
Got it. If I understand correctly then, the title of this PR should be "fix creation of colshapes as children in custom EDFs", right?

Yamsha75:
I agree

jlillis:
Great. I'm going to merge now. I've created a (separate issue)[#262] for adding blips, colshapes, radarareas, and water elements to the editor. Feel free to submit a PR for that issue if you like.

Issue #244, opened 2020-11-08
Map Editor Output Location


umithyo:
**Is your feature request related to a problem? Please describe.**
Yes, I cannot change the output directory of map files.

**Describe the solution you'd like**
Ability to change the location of newly added map files included in settings.xml or something.

qaisjp:
by "output directory" do you mean the organizational directory? e.g. `[newmaps]`?

umithyo:
> by "output directory" do you mean the organizational directory? e.g. `[newmaps]`?

Yeah, exactly.

umithyo:
Can I add this feature and make a PR?

qaisjp:
Sure

jlillis:
Fixed in #245, but discussion continues in #287.

PR #245, opened 2020-11-08
editor: #244 organizational output setting

umithyo:


[editor]/editor_main/meta.xml
```diff
@@ -158,6 +158,12 @@
 			friendlyname="Map backup interval"
 			accept="1-3600"
 			desc="Map backup interval in seconds."
-		/>
+        />
+        <setting name="#mapResourceOrganizationalDirectory"
```
Comment (xLive) [inline [editor]/editor_main/meta.xml:162]: How to change this setting?
because it can't be changed by the admin panel
read about access modifiers:
https://wiki.multitheftauto.com/wiki/Settings_system

Also without a default `value` the setting won't show in the admin panel
https://github.com/multitheftauto/mtasa-resources/blob/c4e453e770067dfa8761e04aba3cf579b37bcd23/%5Badmin%5D/admin/server/admin_settings.lua#L32
so I suggest a value like none or false
```suggestion
        <setting name="*mapResourceOrganizationalDirectory" value="none"
```

[editor]/editor_main/server/saveloadtest_server.lua
```diff
@@ -9,6 +9,7 @@ local dumpInterval = get("dumpSaveInterval") and tonumber(get("dumpSaveInterval"
 local fileTypes = { "script","map","file","config","html" }
 local specialFileTypes = { "script","file","config","html" }
 local DESTROYED_ELEMENT_DIMENSION = getWorkingDimension() + 1
+local mapResourceOrganizationalDirectory = get("mapResourceOrganizationalDirectory");
```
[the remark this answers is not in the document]
Comment (umithyo) [reply [editor]/editor_main/server/saveloadtest_server.lua]: Sorry the old habits :)

[editor]/editor_main/meta.xml
Comment (umithyo) [reply [editor]/editor_main/meta.xml:162]: Does "none" equal to "nil" in lua?

Comment (xLive) [reply [editor]/editor_main/meta.xml:162]: > Does "none" equal to "nil" in lua?

Nope, it's just a string, you need to add that in your code
like other resources like `race` for example
 https://github.com/multitheftauto/mtasa-resources/blob/c4e453e770067dfa8761e04aba3cf579b37bcd23/%5Bgamemodes%5D/%5Brace%5D/race/meta.xml#L113

> you can change the setting in meta.xml. I thought that would be more convenient.

I think it's easier for mappers to be in the admin panel
Thoughts?

Comment (jlillis) [reply [editor]/editor_main/meta.xml:162]: It should be standard practice to make all settings available in the admin panel with proper titles, descriptions, and default values. This is how the settings system is intended to be used.

Comment (umithyo) [reply [editor]/editor_main/meta.xml]: Alright, done

[editor]/editor_main/meta.xml
```diff
@@ -158,6 +158,12 @@
 			friendlyname="Map backup interval"
 			accept="1-3600"
 			desc="Map backup interval in seconds."
-		/>
+        />
+        <setting name="*mapResourceOrganizationalDirectory" value="none"
+			friendlyname="Map resource organizational directory"
+            accept = "*"
```
Comment (jlillis) [inline [editor]/editor_main/meta.xml:164]: Fix indentation

[editor]/editor_main/meta.xml
```diff
@@ -158,6 +158,12 @@
 			friendlyname="Map backup interval"
 			accept="1-3600"
 			desc="Map backup interval in seconds."
-		/>
+        />
+        <setting name="*mapResourceOrganizationalDirectory" value="none"
+			friendlyname="Map resource organizational directory"
+            accept = "*"
+            examples="[newmaps]"
+			desc="Organizational directory for newly added map resources."
```
Comment (jlillis) [inline [editor]/editor_main/meta.xml:166]: You should add a note here that the directory name must be in [brackets].

[editor]/editor_main/server/saveloadtest_server.lua
```diff
@@ -9,6 +9,7 @@ local dumpInterval = get("dumpSaveInterval") and tonumber(get("dumpSaveInterval"
 local fileTypes = { "script","map","file","config","html" }
 local specialFileTypes = { "script","file","config","html" }
 local DESTROYED_ELEMENT_DIMENSION = getWorkingDimension() + 1
+local mapResourceOrganizationalDirectory = get("mapResourceOrganizationalDirectory") ~= "none" and get("mapResourceOrganizationalDirectory") or nil
```
Comment (jlillis) [inline [editor]/editor_main/server/saveloadtest_server.lua:12]: You should consider moving this down to line 360 (just above the call to `createResource`), that way the player can change the organizational directory while the editor is running. Also, you should add some code to validate the organizational directory which needs to be in [brackets]. I don't recall if createResource will add brackets to the directory name or not.

jlillis (changes requested):
Looking good so far, just need to make a few tweaks and this is ready to merge.

[editor]/editor_main/meta.xml
Comment (umithyo) [reply [editor]/editor_main/meta.xml:164]: Indentation looks fine on my editor. How can I fix this?
image

Comment (Woovie) [reply [editor]/editor_main/meta.xml:164]: I see a mix of spaces and tabs. Probably part of the issue.

[editor]/editor_main/server/saveloadtest_server.lua
Comment (umithyo) [reply [editor]/editor_main/server/saveloadtest_server.lua:12]: I tested it, it will not add brackets.
Should I use regex for checking the brackets?

[editor]/editor_main/meta.xml
Comment (jlillis) [reply [editor]/editor_main/meta.xml:164]: Same here, and in the diff viewer on GitHub. We generally use 4 spaces instead of tabs.

[editor]/editor_main/server/saveloadtest_server.lua
Comment (jlillis) [reply [editor]/editor_main/server/saveloadtest_server.lua:12]: Yes, that's what I would do.

[editor]/editor_main/meta.xml
```diff
@@ -158,6 +158,12 @@
 			friendlyname="Map backup interval"
 			accept="1-3600"
 			desc="Map backup interval in seconds."
-		/>
+        />
+        <setting name="*mapResourceOrganizationalDirectory" value="none"
+			friendlyname="Map resource organizational directory"
```
Comment (xLive) [inline [editor]/editor_main/meta.xml:163]: I think we need a shorter name?
image

Comment (qaisjp) [reply [editor]/editor_main/meta.xml:163]: "Map base directory"

[editor]/editor_main/meta.xml
Comment (umithyo) [reply [editor]/editor_main/meta.xml:164]: So, is there anything I can do about this? Put 4 spaces instead of a tab?

[editor]/editor_main/server/saveloadtest_server.lua
Comment (umithyo) [reply [editor]/editor_main/server/saveloadtest_server.lua:12]: What should the user get if they didn't follow the right format? An error message or should I just continue as if the option has never been passed? i.e `createResource(resourceName)`

[editor]/editor_main/meta.xml
Comment (qaisjp) [reply [editor]/editor_main/meta.xml:164]: 1. Click "Spaces: 4" or "Tabs: 4" in the bottom of vscode image
2.  then click "Convert indentation to spaces" image

Comment (qaisjp) [reply [editor]/editor_main/meta.xml:164]: Also, you can view whitespace in the editor by changing this setting to something else image

[editor]/editor_main/server/saveloadtest_server.lua
Comment (jlillis) [reply [editor]/editor_main/server/saveloadtest_server.lua]: A debugscript error would be fine.

[editor]/editor_main/meta.xml
Comment (qaisjp) [reply [editor]/editor_main/meta.xml:164]: Oh, it looks like it previously used tabs instead of spaces. Please could you click "Convert indentation to tabs" instead?

Comment (umithyo) [reply [editor]/editor_main/meta.xml]: Okay, done

[editor]/editor_main/server/saveloadtest_server.lua
Comment (umithyo) [reply [editor]/editor_main/server/saveloadtest_server.lua]: Added the check and error

[editor]/editor_main/meta.xml
Comment (qaisjp) [reply [editor]/editor_main/meta.xml]: you need to do it for [editor]/editor_main/server/saveloadtest_server.lua please

Comment (umithyo) [reply [editor]/editor_main/meta.xml]: Sorry, done

jlillis (approved):

jlillis:
This all seems to work well, but I was wondering if you'd be interested in also modifying the editor save/load GUI to show the organizational directory and maybe allow it to be changed from there. It would be a bit better from a usability standpoint. If you'd rather call this finished I'm also happy to just merge it.

umithyo:
> This all seems to work well, but I was wondering if you'd be interested in also modifying the editor save/load GUI to show the organizational directory and maybe allow it to be changed from there. It would be a bit better from a usability standpoint. If you'd rather call this finished I'm also happy to just merge it.

I think that makes sense to show it in the save/load GUI as well, thanks for the suggestion. I will work on it as soon as I'm available.

xLuxy:
What about backups?
https://github.com/multitheftauto/mtasa-resources/blob/e51ed2fb1736c547ca0c7984d43a161d2c6872ac/%5Beditor%5D/editor_main/server/save_backup.lua#L18

jlillis:
> 
> What about backups?

What about them?

umithyo:
I don't think I will be able to do add these settings into save/load GUI for a while. Sorry for the long wait. Can we merge this and look into it later?

jlillis:
Sure, I'll merge this for now and open a feature request for that separately.

PR #247, opened 2020-11-10
Add .editorconfig file

jlillis:
When merged, this will add an [.editorconfig](https://editorconfig.org/) file to the repo to help enforce a consistent coding style. The following rules are set:

- Add final newline and trim trim trailing whitespace from all files
- Use 4 tabs instead of spaces and crlf line endings in .lua files

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# top-most EditorConfig file
+root = true
+
+[*]
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.lua]
+indent_style = space
+indent_size = 4
+end_of_line = crlf


qaisjp (approved):
I think in the long term we'll want to standardise on LF (as we have in mtasa-blue), but for now, since CRLF is the most common, we can stick with this.

In the future, we should have some sort of formatter that enforces all Lua formatting. We can change all CRLF to LF in one fell sloop, along with the Lua formatting enforcer.

----

Extra context about formatting: there are/were dreams of adding a Lua formatter that will enforce all Lua code to be formatted a certain way.  We just needed to finish off https://github.com/trixnz/lua-fmt/pull/21 and then apply it to this repo.  Time and priorities meant that I never got round to this. Also, that project was abandoned even when I commented on that PR. That said, we applied lua-fmt to the admin2 resource, as a proof of concept.

We didn't apply it to the entire repository as we wanted to make sure we do only one mass formatting commit, instead of multiple over the years. (Unfortunately, it looks like this commit 2c6416a09697b9d8b812c182cff0955b09d753c5 slipped through.)

jlillis:
Did commit https://github.com/multitheftauto/mtasa-resources/commit/2c6416a09697b9d8b812c182cff0955b09d753c5 standardize the repo on LF or CRLF? If LF, then I should update my settings here I think. Also, is there any way to perform a "simpler" format checking at just checks for proper indentation and line endings? That seems to be the most common pain point.

qaisjp:
Unfortunately I think it accidentally MISstandardised it to CRLF. It seems to edit all admin2 files, which were standardised to LF by lua-fmt.

qaisjp:
> Also, is there any way to perform a "simpler" format checking at just checks for proper indentation and line endings?

I think we could do that. Do you want to look into that, or shall I?

jlillis:
>I think we could do that. Do you want to look into that, or shall I?
I have zero experience with GitHub Actions/CI so it's probably best if you look into it.

Looks like you're correct: `.gitattributes` calls for LF line endings. 

If we can get consensus on the settings I've defined so far I'll go ahead and add a CONTRIBUTING.md file describing the desired coding style so far.

A seperate PR to format existing code is probably called for too.

qaisjp:
Oh, hm, if our `.gitattributes` calls for LF, and most of our current code is CRLF, I'm not sure what to do.

Maybe we can just comment out that line of the `.gitattributes` file, and revisit line endings (and mass-formatting) later.

jlillis:
I think that's probably the best move for now. I'll open a separate issue for discussion.

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+root = true
+
+[*]
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.lua]
+indent_style = space
+indent_size = 4
+end_of_line = crlf
```
qaisjp, line 11: Maybe we should comment this out too?  The repo is currently mixed, right?

jlillis:
Going to switch this back to a draft PR until we make some more decisions in #250

Fernando-A-Rocha:
I think it's safe to implement this now

shadylua:
have you tried `npx prettier --write .` to format the codes??

lynconsl:
Any chance of it being implemented?

PR #254, opened 2020-11-23
race: end the map when only spectators are left

### LINKED ISSUE #191: (Race Gamemode) End map when only spectators are left
The race state is normally set to "EveryoneFinished" whenever... well everyone finishes a race. This skips the "timeafterfirstfinish" waiting time and starts a vote for the next map. However this race state is never set when there are spectators still left watching the race after all non-spectating players have finished. This makes it quite frustrating to have any spectators.

Set the race state to "EveryoneFinished" when all non-spectating players have finished, as opposed to every single player currently on the server. Perhaps making a new setting to configure whether you'd like to skip waiting for spectators or not.

It is possible to just lower the "timeafterfirstfinish" amount to lower the wait, however then you have issues with people who are trying to race not finishing in time. For longer races you want a higher "timeafterfirstfinish" but for shorter ones waiting for say a minute after a 20 second race for some spectators slows things down a lot.

**Additional context**
I was planning on adding an afk script to a race server whereby afk players are forced into spectator mode as opposed to just kicking them from the server (say for instance they crashed out early and don't want to continue, or don't like the map and would like to just watch) however with spectators holding the races up it's not worth doing.

xLive:
Fixes #191

### DIFF
--- a/[gamemodes]/[race]/race/meta.xml
+++ b/[gamemodes]/[race]/race/meta.xml
@@ -297,6 +297,12 @@
 					accept="false,true"
 					desc="Force players to spawn like after death."
 					/>
+		<setting name="*endmapwhenonlyspectators" value="true"
+					friendlyname="End map when only spectators left"
+					group="Spectating"
+					accept="false,true"
+					desc="Set to true to end the map when only spectators are left"
+					/>
 
 		<setting name="*joinrandomvote" value="true"
 					friendlyname="Random vote on join"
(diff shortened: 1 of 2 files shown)


jlillis (commented):
This code looks fine. I don't have a server with more than one player (myself) to test it on - could another reviewer try this out?

jlillis (changes requested):
This PR looks good but currently contains merge conflicts. Once you resolve this, I'd be happy to merge.

xLive:
done

jlillis (approved):

PR #256, opened 2020-11-24
editor: add ability to lock elements

### LINKED ISSUE #238: Add ability to lock objects in Map Editor
When I have a large ground object placed, I sometimes accidentally click on it and move it.

It would be really nice to have an option to 'lock' objects, making them unclickable. This could be very useful for interior mappers.

xLive:
Closes #238 

To lock/unlock an element use the 'L' key
you still can select it from the elements browser

### DIFF
--- a/[editor]/editor_main/client/main.lua
+++ b/[editor]/editor_main/client/main.lua
@@ -58,6 +58,8 @@ local DRAG_CAMERA_MINIMAL_DISTANCE = 2
 local g_lastClick = { tick=getTickCount() }
 local DOUBLE_CLICK_MAX_DELAY = 500 -- In ticks
 
+local g_lock = {}
+
 setCloudsEnabled(false) -- We don't need clouds.
 setAmbientSoundEnabled("gunfire", false) -- We don't need random gun shots.
 
@@ -485,6 +487,11 @@ function processClick ( clickedElement, key, keyState, lookX, lookY, lookZ )
 			end
 		end
 
+		if g_selectedElement ~= clickedElement and g_lock[clickedElement] then
+			editor_gui.outputMessage("This element is locked, Press '"..cc["lock_selected_element"]:upper().."' to unlock it.", 255,255,255)
+			return false
+		end
+
 		if (key == "select_target_mouse") then
 			selectElement(clickedElement, MOUSE_SUBMODE, false, g_selectedElement, g_selectedElement)
 		elseif (key == "select_target_keyboard") then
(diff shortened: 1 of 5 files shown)


[editor]/editor_main/client/main.lua
```diff
@@ -987,6 +1000,26 @@ function cloneSelectedElement()
...
+function lockSelectedElement()
+	local targetElement = getTargetedElement()
+	if targetElement then
+		if g_lock[targetElement] then
+			g_lock[targetElement] = nil
+			editor_gui.outputMessage("You have unlocked this element.", 0,255,0)
+		else
+			g_lock[targetElement] = true
+			editor_gui.outputMessage("You have locked this element.", 255,0,0)
+		end
```
Comment (qaisjp) [inline [editor]/editor_main/client/main.lua:1012]: only errors should use red

Comment (xLive) [reply [editor]/editor_main/client/main.lua:1012]: > i think only errors should use red

so what do you suggest? green for both of them?

Comment (qaisjp) [reply [editor]/editor_main/client/main.lua:1012]: i think yellow is frequently used (255, 255, 0) - https://github.com/multitheftauto/mtasa-resources/search?q=outputMessage

Comment (xLive) [reply [editor]/editor_main/client/main.lua:1012]: most messages with yellow colors are warnings

Comment (qaisjp) [reply [editor]/editor_main/client/main.lua:1012]: I'll leave it to you decide what colour to use, but this is what I found:

**alerts rgb(255, 255, 0)**

image image image

**i don't know what colours these are**

image image

----

github code search is pretty bad though, so maybe you have a clearer view when doing a full text search in vscode

Comment (xLive) [reply [editor]/editor_main/client/main.lua:1012]: this is what I found:

- Red: for errors and main actions like saving/opening a map
- Yellow: for alerts as you said
- White: when you cannot select the element
- Green: for a successful action
only used for one message currently:
https://github.com/multitheftauto/mtasa-resources/blob/6eeaea3c679c68d530e8d0e162657d6200c7035c/%5Beditor%5D/editor_gui/client/browser/browser.lua#L392

so I think green is the perfect color here?

Comment (qaisjp) [reply [editor]/editor_main/client/main.lua:1012]: cool, go for it :)

Comment (xLive) [reply [editor]/editor_main/client/main.lua]: done

jlillis (approved):
Looks good. Only thing I would consider adding is some sort of lock icon that an element in the element browser and the element's properties window that indicates that the element is locked and allows the user to lock the element from there. If you're interested in doing this I'll wait to merge, otherwise I'm happy to merge now.

xLive:
> 
> 
> Looks good. Only thing I would consider adding is some sort of lock icon that an element in the element browser and the element's properties window that indicates that the element is locked and allows the user to lock the element from there. If you're interested in doing this I'll wait to merge, otherwise I'm happy to merge now.

I decided to add it without icons because I don't have one and gridlist/properties window doesn't support images anyway,
you can now lock the element from the properties window
and check if it's locked from the element browser
mta-screen_2020-12-06_22-30-25
mta-screen_2020-12-06_22-30-59
What do you think?

jlillis (approved):

PR #257, opened 2020-11-30
editor: Change `usedLODModels` structure

iDannz-Breno:
This pull request changes the table structure that transfers the lod models to the client to apply `engineSetModelLODDistance`, which can make loading much faster for resources with very large maps and with many LODS of the same ID.

Currently if the resource has the `useLODs` setting enabled in `meta.xml`, it loop all objects checking if that object has a valid LOD model (using a pre-written table inside the script), and if it does, it will add that model LOD within a table with `table.insert`.

The problem is: if the map file is very large and has many repeated objects with the same lod model, this value will be added several times within the table, and this can make the table very large, when in fact this lod model would only need to be stored once.


qaisjp (approved):
Nice find.

PR #259, opened 2020-12-05
admin2: automatic scripts

### LINKED ISSUE #102: admin2: automatic scripts
**Describe the bug**
"Automatic scripts" can be seen on the "Server" tab but you cannot edit the fields.

**To do**
1. Investigate the existing code
2. Link up the GUI with the existing code OR implement the features

**Expected behavior**
To be able to edit the fields

**Screenshots**
image

iDannz1:
This pull request closes #102.

### DIFF
--- a/[admin]/admin2/client/main/admin_automatic-scripts.lua
+++ b/[admin]/admin2/client/main/admin_automatic-scripts.lua
@@ -54,27 +54,6 @@ aAutoScripts["fpskicker"] = {
     end
 }
 
-aAutoScripts["fpskicker"] = {
-    state = nil,
-    timer = nil,
-    limit = 10,
-    checkInterval = 3000,
-    toggle = function(state)
-        if isElement(aServerTab.FPSKickerCheck) then
-            guiCheckBoxSetSelected(aServerTab.FPSKickerCheck, state)
-        end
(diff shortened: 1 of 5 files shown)


jlillis:
I have a few concerns with this implementation:

1: we are relying on the client to perform the checks. A malicious client could simply not perform the checks (i.e fail to load the script). I recommend performing these checks server-side and just polling the client for needed information (i.e fps, idletime - ping can be determined serverside with [getPlayerPing](https://wiki.multitheftauto.com/wiki/GetPlayerPing).
2: values for minimum fps and maximum idletime and ping should be used as average values rather than absolute values - this will allow for some wiggle-room in the case of temporary network/pc disruptions. The degree of wiggle-room should be configurable via settings. Say we want to kick players who average under 15fps for 60 seconds: minimumFPS = 15fps, period = 60seconds.
3: the term "automatic scripts" seems vague to me. I would recommend splitting this into two different scripts: "latency checks" for fps/ping, and "idle timeout" for idletime.

iDannz1:
Lol, sorry for the delay, I totally forgot about this PR and also thought I had already left an answer.

I think another PR is more reasonable than applying these changes since a lot of code needs to be changed. If no one picks it, maybe another day I do it again, following your ideas.

PR #263, opened 2020-12-06
Add editor_dump,editor_test to gitignore

xLive:


.gitignore
```diff
@@ -15,3 +15,7 @@
 
 # acpanel
 \[admin\]/acpanel/settings.xml
+
+# editor
+\editor_dump/
+\editor_test/
```
qaisjp, line 21:
```suggestion
editor_dump/
editor_test/

```
The backslash was escaping `[` it's not required at the start of every line

xLive: done

jlillis (approved):

PR #264, opened 2020-12-14
Make collisionless world models selectable in map editor

### LINKED ISSUE #251: Editor - Select collisionless
It's not possible to select objects that has no collision model. E.g bushes and night lights.

Add an option in Editor to select based on geometry or element(?)

Add an option in Editor to load collision patches

Option B - I'm able to get collisions ready in 1-2 weeks. I'll need someone to implement it into Editor and add a button or shortcut. 
Size is 148 bytes/patch

Zangomangu:
Fixes #251 

This PR makes it possible to select and remove world models that were previously collisionless.

- Loads in 1.288 collision files (196 kB) from 'patches.col'. Credit and applause to @tut98 who manually created all these collision files, quite a job. They are stored in a col archive because there are well known tools to edit these, and they have no overhead.
- From a checkbox in editor settings, these can be toggled on, which does the following:
- - Col elements are created and replaced.
- -  4116 "ghost" objects are created in places of world models. These are defined in 'placement.list' which is compiled from IPL entries that contain affected models. This is to force GTA to load in collision, many but not all affected models require this.
- - Loading takes ~5 seconds at 36 FPS. It could be made slower or faster, this is a compromise between smooth loading and speed that also considers the main playerbase on a pentium III
- When the setting is toggled off
- - Col models are restored and col elements destroyed
- - Objects are destroyed

The setting is automatically toggled off when going into test mode, and back on when leaving, because it makes testing unplayable.


jlillis:
The code looks good and I'll be testing this shortly but here's a question: can you think of any scenarios where these collision fixes would be useful outside of the editor? If so, should we consider putting them in a separate resource?

Zangomangu:
> The code looks good and I'll be testing this shortly but here's a question: can you think of any scenarios where these collision fixes would be useful outside of the editor? If so, should we consider putting them in a separate resource?

When you load them in, you can barely move around the map outside editor mode. They add col to shadows, power lines, vegetation, and other mainly visual stuff that isn't meant to have collision, so I can't think of any useful scenario. Some of them are for example a simple box because the model contains many smaller parts.

jlillis (approved):
Tested and works as intended. 👍

PR #266, opened 2020-12-24
resourcemanager: Fix start/load dates

### LINKED ISSUE #265: [resourcemanager] Start/load dates are incorrect

[web]/resourcemanager

**Describe the bug**

Resource start and load dates are one month behind.

**To Reproduce**
1. Go to web admin
2. Click on Resource manager
3. Click on any resource
4. Observe start and load dates

**Expected behavior**

Dates should be represented correctly

**Screenshots**

Resource was started on 2020-12-24, but the date is  2020-11-24
image

Dezash:
Fixes #265

### DIFF
--- a/[web]/resourcemanager/manager.lua
+++ b/[web]/resourcemanager/manager.lua
@@ -22,21 +22,9 @@ function getResourceInfo ( resource )
 
     local failreason =  getResourceLoadFailureReason ( resource )
     local state = getResourceState ( resource )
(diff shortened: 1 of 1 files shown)


xLive (changes requested):
This will cause an error if the resource hasn't been started yet,
it should show "Never" like before

Dezash:
> This will cause an error if the resource hasn't been started yet,
> it should show "Never" like before

Fixed

jlillis (approved):
👍

PR #271, opened 2021-01-18
Race : Problem to restore rotation of vehicles - restorePlayer function - base.lua

Mkl21:
Hello,

While playing destruction derby maps with respawn of vehicles, rotation of vehicles had a little problem while respawning :

`WARNING: [gamemodes]/[race]/race/modes/base.lua:438: Bad argument @ 'setVehicleRotation' [Expected number at argument 4, got table]`

Here I suggest the way I solved it.

Have a nice evening.

### DIFF
--- a/[gamemodes]/[race]/race/modes/base.lua
+++ b/[gamemodes]/[race]/race/modes/base.lua
@@ -417,7 +417,7 @@ function restorePlayer(id, player, bNoFade, bDontFix)
 	if not RaceMode.checkpointsExist() or checkpoint==1 then
 		local spawnpoint = self:pickFreeSpawnpoint(player)
 		bkp.position = spawnpoint.position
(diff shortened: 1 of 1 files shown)


[gamemodes]/[race]/race/modes/base.lua
```diff
@@ -417,7 +417,7 @@ function restorePlayer(id, player, bNoFade, bDontFix)
 	if not RaceMode.checkpointsExist() or checkpoint==1 then
 		local spawnpoint = self:pickFreeSpawnpoint(player)
 		bkp.position = spawnpoint.position
-		bkp.rotation = {0, 0, spawnpoint.rotation}
+		bkp.rotation = {0, 0, spawnpoint.rotation[3]}
```
Comment (jlillis) [inline [gamemodes]/[race]/race/modes/base.lua:420]: Would it be better to do `bkp.rotation = spawnpoint.rotation`?

jlillis (approved):

PR #272, opened 2021-01-20
Use predefined MTA variables

xLuxy:
This PR replaces stuff like `getResourceRootElement ( getRootElement( ) )` with pre-defined MTA variables found here: https://wiki.multitheftauto.com/wiki/Predefined_variables_list


ArranTuna:
I've done all the root, localPlayer and resourceRoot.

PR #274, opened 2021-01-22
stop trying to restart non-running resources

Dante383:
trying to restart non-running resource results in a warning in the debug console


jlillis (changes requested):
Looks like you added a commit intended for another PR. Revert the ipb commit and I can merge this.

Dante383:
I'm just gonna create a new PR because I messed a bit with the git config since last commit and it'd take way too long to fix it

PR #276, opened 2021-01-23
Create README for ipb resource

Dante383:


jlillis:
We currently use the wiki for documentation. See https://wiki.multitheftauto.com/wiki/Category:Resource

Issue #250, opened 2020-11-13
Establish & enforce style rules


jlillis:
I think it would be a good idea to establish and begin enforcing (via code reviews and CI checks) style rules for this repository just like we do for mtasa-blue. To this end, I have [submitted a PR](https://github.com/multitheftauto/mtasa-resources/pull/247) that adds a .editorconfig file that defines the following basic rules:

- No trailing whitespace
- Insert final newlines
- Use 4 spaces instead of tabs
- Use CRLF line endings

## Line endings
There has been some inconsistency regarding which type of line endings are meant to be used. The .gitattributes file calls for LF endings and [it's been mentioned](https://github.com/multitheftauto/mtasa-resources/pull/247#pullrequestreview-527673060) that in the long-term we want LF, but a [recent commit](https://github.com/multitheftauto/mtasa-resources/commit/2c6416a09697b9d8b812c182cff0955b09d753c5) seems to have standardized on CRLF. I guess deciding on this would be the first order of buisness.

## CI checks
On CI, there have been previous attempts to [use lua-fmt](https://github.com/multitheftauto/mtasa-resources/tree/scratch-luafmt) to automate formatting checks. It looks like these efforts have stagnated. Perhaps it would be possible to either a) restart work on lua-fmt or b) do something else? I was thinking perhaps we can just run a simple CI check for the style rules I've described above.

## Mass-edits and ignore-revs
Once we've decided on a set of rules and implemented the controls to enforce them, another mass-formatting edit will be needed. We should add a .git-blame-ignore-revs file to filter this and previous such commits out.

## TLDR

- [ ] establish basic style rules and document them
- [ ] update .editorconfig and .gitattributes accordingly
- [ ] setup related CI checks
- [ ] issue mass-format commit
- [ ] setup .git-blame-ignore-revs

jlillis:
I have a new rule to propose:
- Use the global variables `root`, `resource`, and `resourceRoot` rather than calling `getRootElement()`, `getResource()`, `getResourceRootElement(resource)`

patrikjuvonen:
For additional reference I had worked on a template for styles at https://github.com/multitheftauto/mtasa-resources/wiki/Lua-style-guide

jlillis:
To get this moving again, can we decide on which line endings are the right line endings? I assume we'd want to use the same style as mtasa-blue, which is CRLF right?

patrikjuvonen:
Specific style enforcement is required for JS and CSS.

[...]

PR #281, opened 2021-01-24
Upgrade and fix webmap

patrikjuvonen:
## Summary
- Changes ajax to use Object.defineProperty to avoid direct prototype manipulation and to disable enumeration (required for OpenLayers to work)
- Switch the OpenLayers layer to a working map asset link (assets.mtasa.com)
  - Reason for hosting the files instead of putting them in here is because of the huge file size of the tiles
- Change background color of the webmap to be sea color, it's better for eyes instead of the white background
- Update OpenLayers to 6.5.0
- Add support for radar blip id 0 with color and size
- Dead players are now shown as a red tinted player blip
- Player blips support player rotation now as well
- Add current map coordinates from mouse position to the top right corner
- Refactored code a little bit

Untitled


eAi:
Hi @patrikjuvonen - I may have those assets for the map around somewhere still. I’ll try to dig them out tomorrow.

patrikjuvonen:
> Hi @patrikjuvonen - I may have those assets for the map around somewhere still. I’ll try to dig them out tomorrow.

Great, thank you!

eAi:
Here’s the image that was used for the tilecache: 



It’s about 68MB.

I’ve got the original tilecache cgi (and pythons scripts) that must have run this, but I suspect there’s a much better way to do this these days. You’re welcome to them if it’s useful.

Given that this map is relatively small, it would seem that generating a static set of tiles would be the best approach. I found this script that claims to do this: https://github.com/jahed/maptiles

patrikjuvonen:
> Here’s the image that was used for the tilecache:
> 
> http://www.opencoding.net/misc/gtasa_aerial_map.PNG
> 
> It’s about 68MB.
> 
> I’ve got the original tilecache cgi (and pythons scripts) that must have run this, but I suspect there’s a much better way to do this these days. You’re welcome to them if it’s useful.
> 
> Given that this map is relatively small, it would seem that generating a static set of tiles would be the best approach. I found this script that claims to do this: https://github.com/jahed/maptiles

Thanks! I'll generate new tiles from this image, should be all I need 😃

patrikjuvonen:
I managed to generate the tiles successfully.

@eAi Does either of your scripts contain the tile size used before? Max zoom level appears to have been 6.

eAi:
The tiles were dynamically generated before, it looks like it defaulted to generating 20 levels. It generated them using https://github.com/OSGeo/tilecache

patrikjuvonen:
> The tiles were dynamically generated before, it looks like it defaulted to generating 20 levels. It generated them using https://github.com/OSGeo/tilecache

I see, thanks!

patrikjuvonen:
New changes in this PR:
- Update OpenLayers to 6.5.0
- Add support for radar blip id 0 with color and size
- Dead players are now shown as a red tinted player blip
- Player blips support player rotation now as well
- Refactored code a little bit

[web]/webmap/.editorconfig
```diff
@@ -0,0 +1,22 @@
+# This file is for unifying the coding style for different editors and IDEs
```
Comment (jlillis) [inline [web]/webmap/.editorconfig:1]: Please remove this file, we will be using a single .editorconfig file once agreement is reached on what settings it will contain (#250).

[web]/webmap/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
Comment (jlillis) [inline [web]/webmap/.prettierrc:1]: Same as above; please remove this file.

[web]/webmap/meta.xml
```diff
@@ -1,14 +1,17 @@
 <meta>
-    <include resource="ajax" />
-    <script src="map.lua" />
-	<html src="map.htm" default="true"/>
-    <html src="css.css" raw="true" />
-    <html src="script.js" raw="true" />
+	<info author="eAi" name="Map" showInResourceBrowser="true" version="2.0.0" />
```
Comment (jlillis) [inline [web]/webmap/meta.xml:2]: Feel free to add yourself as an author, especially if you plan on working on this and/or related web resources in the future. Also, I think "Web Map" may have been a better name but it's up to you.

jlillis (dismissed):
Looks good overall, but there are some editorconfig/prettierrc files added that shouldn't be there. We are still discussing style rules for the repo in #250

[web]/webmap/meta.xml
Comment (patrikjuvonen) [reply [web]/webmap/meta.xml:2]: I decided to leave it to Map since the admin interface side navigation depends on "Map" name and I couldn't be bothered to switch it in the other resource.

jlillis (approved):
lgtm

PR #282, opened 2021-01-25
Lang support for editor resource

fresholia:
By editing JSON files, members of country communities can easily integrate systems into their own languages.
The resource is still under development.

Do not translate placeholders like %s, they will be auto-replaced by the script.

Like this:

JSON:
```json
"TRANSLATE_TXT": "result is: %s"
```

Lua:
```lua
exports.editor_lang:_("TRANSLATE_TXT"):format("aaa")
-- print result is: aaa
```


jlillis:
I don't think using a separate resource for editor translations and JSON files is an ideal implementation. We should make use of the existing infrastructure at https://translate.mtasa.com/ which is based on the GNU gettext standard.

sbx320:
I think the cleaner way is to add a generic translation resource, which accepts standard `.po` files via exports and then offers a translation service back.

turret001:
> I don't think using a separate resource for editor translations and JSON files is an ideal implementation. We should make use of the existing infrastructure at https://translate.mtasa.com/ which is based on the [GNU gettext standard](https://www.gnu.org/software/gettext/).

I wouldn't trust translate.mtasa.com on this, seeing how bad some Turkish translations were done

cleopatradev:
> I think the cleaner way is to add a generic translation resource, which accepts standard `.po` files via exports and then offers a translation service back. That way it'd also be useful for other non-default resources.

It would make sense to make a single language resource for all resources, but we should do this with JSON.

xLive:
> > I don't think using a separate resource for editor translations and JSON files is an ideal implementation. We should make use of the existing infrastructure at https://translate.mtasa.com/ which is based on the [GNU gettext standard](https://www.gnu.org/software/gettext/).
> 
> I wouldn't trust translate.mtasa.com on this, seeing how bad some Turkish translations were done

You can always review and suggest the translations there.

Fernando-A-Rocha:
It makes sense to have a central resource for all translations, not just one for the Map Editor.

I've no idea how we would fetch the translations from Crowdin, but I think it would be better in the long run than storing strings in JSON files. Those would be harder to maintain because they'd require PRs to this repository.

Fernando-A-Rocha:
#361 is related

jlillis:
Given the age and lack of progress here, I'm going to close this PR for now.

PR #283, opened 2021-01-25
Fix #280

Dante383:

Issue #280, opened 2021-01-24
Fresh installation has errors


ciber96:
**Which resource(s) have this problem?**
admin

**Describe the bug**
Fresh installation of admin resources has sevral errors not present before and spams the console.

**To Reproduce**
1. Fresh install of Admin Resource

**Expected behavior**
No errors

**Screenshots**
https://i.ciber.dev/s/jWPBTgqBqgcnfpC/preview
https://i.ciber.dev/s/X7fFzbsWetDBtJ4/preview
https://i.ciber.dev/s/n6maxcGXDMNKbxb/preview

**Version**
Not aplicable

**Additional context**
Latest Version available from linux.mtasa.com

sbx320:
Looks like there's an `end` too much here:

https://github.com/multitheftauto/mtasa-resources/commit/d237f279a5e6508864c296798a65c55b69139dd6#diff-cff77698f0b2306e72e28e9c7ded09859eaded7d1987c9c606a79fe7e9989dceR1268

We should have some sort of Github action to check if comitted files are valid Lua

jlillis:
Fixed by #284

PR #284, opened 2021-01-26
admin: it updates admin_server.lua and admin_sync.lua

andr0xy:
admin_server.lua: an end with no need.
admin_sync.lua: sometimes aPlayers table doesn't get the correct index in loop. Added isElement function to check.


xLive:
#283

xLive:
About `admin_sync` file:
`aPlayers` table is nil because `admin_server` has errors.

jlillis (approved):
This fixes the errors occuring on master as identified by #280. Thanks!

PR #285, opened 2021-01-26
admin: fix output message on actions

andr0xy:
admin

It happens when a player has % after his nickname, the output message isn't complete.
That PR fixes it.

### DIFF
--- a/[admin]/admin/server/admin_server.lua
+++ b/[admin]/admin/server/admin_server.lua
@@ -524,11 +524,11 @@ function aAction ( type, action, admin, player, data, more )
 	if ( aLogMessages[type] ) then
 		function aStripString ( string )
 			string = tostring ( string )
-			string = string.gsub ( string, "$admin", getPlayerName ( admin ) )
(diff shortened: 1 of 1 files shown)


[admin]/admin/server/admin_server.lua
```diff
@@ -522,13 +522,20 @@ addCommandHandler ( "admin", aAdminMenu )
 
 function aAction ( type, action, admin, player, data, more )
 	if ( aLogMessages[type] ) then
+		function aEscapeNickname( name )
```
Comment (xLuxy) [inline [admin]/admin/server/admin_server.lua:525]: Make this a local function (otherwise you're defining it on the global scope in which case you can put that function somewhere else - which is unnecessary in our case)

[admin]/admin/server/admin_server.lua
```diff
@@ -522,13 +522,20 @@ addCommandHandler ( "admin", aAdminMenu )
 
 function aAction ( type, action, admin, player, data, more )
 	if ( aLogMessages[type] ) then
+		function aEscapeNickname( name )
+			if string.find( name, "%%" ) then
```
Comment (xLuxy) [inline [admin]/admin/server/admin_server.lua:526]: There's no need for this if statement to be here

xLuxy (changes requested):

Dutchman101 (approved):

PR #286, opened 2021-01-27
Add higher quality image to freeroam

patrikjuvonen:
The current freeroam map image is 600x600px and compressed heavily. I can't make out Grove Street from the image which shouldn't be the case.

This PR updates the image to a compressed 1000x1000px image. File size increase is around 2 times (roughly 200KB -> 400KB), which should be fair enough.


Citizen01:
Totally fair 👍

Citizen01 (approved):

jlillis (approved):

PR #290, opened 2021-02-13
headshot: minor improvements

majqqq:
👋🏻
- Removed triggerEvent, in fact there's no need for it, since no other resource use it.
- Replaced getRootElement() with root predefined variable.
- Switched to non-anonymous functions.

### DIFF
--- a/[gameplay]/headshot/headshot.lua
+++ b/[gameplay]/headshot/headshot.lua
@@ -1,26 +1,19 @@
-local removeHeadOnHeadshot = get("removeHeadOnHeadshot");
+local removeHeadOnHeadshot = get("removeHeadOnHeadshot")
 
-addEvent "onPlayerHeadshot"
+function onPlayerDamage(attacker, weapon, bodypart, loss)
+	if bodypart == 9 then
+		killPed(source, attacker, weapon, bodypart)
 
-addEventHandler("onPlayerDamage", getRootElement(),
-	function (attacker, weapon, bodypart, loss)
-		if bodypart == 9 then
-			local result = triggerEvent("onPlayerHeadshot", source, attacker, weapon, loss)
-			if result == true then
-				killPed(source, attacker, weapon, bodypart)
-				if removeHeadOnHeadshot then
-					setPedHeadless(source, true)
-				end
-			end
+		if removeHeadOnHeadshot then
+			setPedHeadless(source, true)
 		end
 	end
-)
+end
+addEventHandler("onPlayerDamage", root, onPlayerDamage)
 
-addEventHandler("onPlayerSpawn", getRootElement(),
-	function()
-		-- Restore head if it got blown off
-		if removeHeadOnHeadshot and isPedHeadless(source) then
-			setPedHeadless(source, false)
-		end
+function onPlayerSpawn()
+	if removeHeadOnHeadshot and isPedHeadless(source) then
+		setPedHeadless(source, false) -- Restore head if it got blown off
 	end
-)
+end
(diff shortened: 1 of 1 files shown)


jlillis (changes requested):
Removing the onPlayerHeadshot API event will break other resources that depend upon it. If you reverse that change, we can then approve the rest of this PR.

majqqq:
My bad, fixed :)

jlillis (approved):

[gameplay]/headshot/headshot.lua
```diff
@@ -1,26 +1,25 @@
...
+end
+addEventHandler("onPlayerDamage", root, onPlayerDamage)
 
-addEventHandler("onPlayerSpawn", getRootElement(),
-	function()
-		-- Restore head if it got blown off
-		if removeHeadOnHeadshot and isPedHeadless(source) then
-			setPedHeadless(source, false)
-		end
+function onPlayerSpawn()
```
Comment (Dezash) [inline [gameplay]/headshot/headshot.lua:20]: Function name could be more clear, how about **restoreHead**?

[gameplay]/headshot/headshot.lua
```diff
@@ -1,26 +1,25 @@
...
-addEventHandler("onPlayerDamage", getRootElement(),
-	function (attacker, weapon, bodypart, loss)
-		if bodypart == 9 then
-			local result = triggerEvent("onPlayerHeadshot", source, attacker, weapon, loss)
-			if result == true then
-				killPed(source, attacker, weapon, bodypart)
-				if removeHeadOnHeadshot then
-					setPedHeadless(source, true)
-				end
+function onPlayerDamage(attacker, weapon, bodypart, loss)
```
Comment (Dezash) [inline [gameplay]/headshot/headshot.lua:5]: How about renaming this function to **headshotCheck**?

[gameplay]/headshot/headshot.lua
Comment (majqqq) [reply [gameplay]/headshot/headshot.lua:20]: Since it is a binded to event, i suggest: **onPlayerSpawnRestoreHead**

[gameplay]/headshot/headshot.lua
Comment (majqqq) [reply [gameplay]/headshot/headshot.lua:5]: According to what i've wrote above, maybe: **onPlayerDamageHeadshot**

Comment (Dezash) [reply [gameplay]/headshot/headshot.lua:5]: **handleDamage**?

[gameplay]/headshot/headshot.lua
Comment (Dezash) [reply [gameplay]/headshot/headshot.lua:20]: In that case, I would go with **handleSpawn**

Comment (jlillis) [reply [gameplay]/headshot/headshot.lua:20]: I think your original function name was acceptable, but if you want something more "ideal" I recommend a name that simply indicates what the function does: `restorePlayerHead` would be my first choice.

Also, you could make this function local.

[gameplay]/headshot/headshot.lua
Comment (jlillis) [reply [gameplay]/headshot/headshot.lua:5]: As mentioned above the "ideal" name might be "checkForHeadshot" - but your original name is acceptable too. Again, this function could be local.

Comment (majqqq) [reply [gameplay]/headshot/headshot.lua]: > this function could be local.
Do you want to enforce local functions everywhere where it is possible or just for events? Would be nice to know when i'll work on other resources :)

Comment (jlillis) [reply [gameplay]/headshot/headshot.lua]: It's not a hard requirement, it's just considered good practice to do where possible. Local functions are slightly faster than global ones.

Comment (majqqq) [reply [gameplay]/headshot/headshot.lua]: Yeah i'm aware of it, thanks for answer.

jlillis (approved):

PR #292, opened 2021-02-13
reload: few changes

majqqq:
Client:
- Replaced anonymous function with non anonymous, it's called reloadWeapon now.
- Renamed function **reloadWeapon** -> **reloadTimer** since it's called via timer.
- Due of that reloadWeapon it's now used as command handler/bind.
- triggerServerEvent uses localPlayer instead of resourceRoot.

Server:
- Localized event function.
- Added if check for client, according [to this wiki article.](https://wiki.multitheftauto.com/wiki/Script_security)
- Event's bound element changed to root, so it could also accept player as a source element.

### DIFF
--- a/[gameplay]/reload/reload.lua
+++ b/[gameplay]/reload/reload.lua
@@ -1,5 +1,7 @@
-function reloadWeapon()
-	reloadPedWeapon(client)
+local function reloadWeapon()
+	if client then
+		reloadPedWeapon(client)
+	end
 end
 addEvent("relWep", true)
-addEventHandler("relWep", resourceRoot, reloadWeapon)
+addEventHandler("relWep", root, reloadWeapon)
--- a/[gameplay]/reload/reload.lua
+++ b/[gameplay]/reload/reload.lua
@@ -1,7 +1,5 @@
 local function reloadWeapon()
-	if client then
-		reloadPedWeapon(client)
-	end
+	reloadPedWeapon(client)
 end
 addEvent("relWep", true)
 addEventHandler("relWep", root, reloadWeapon)
--- a/[gameplay]/reload/reload_c.lua
+++ b/[gameplay]/reload/reload_c.lua
@@ -13,17 +13,20 @@ local blockedTasks = {
     ["TASK_SIMPLE_CAR_OPEN_DOOR_FROM_OUTSIDE"] = true,
 }
 
-local function reloadWeapon()
+local function reloadTimer()
     if blockedTasks[getPedSimplestTask(localPlayer)] then
         return
     end
-    triggerServerEvent("relWep", resourceRoot)
+
+    triggerServerEvent("relWep", localPlayer)
 end
 
 -- The jump task is not instantly detectable and bindKey works quicker than getControlState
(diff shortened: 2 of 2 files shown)


[gameplay]/reload/reload.lua
```diff
@@ -1,5 +1,7 @@
-function reloadWeapon()
-	reloadPedWeapon(client)
+local function reloadWeapon()
+	if client then
```
Comment (jlillis) [inline [gameplay]/reload/reload.lua:2]: This check is not needed; `client` is guaranteed to be valid.

[gameplay]/reload/reload.lua
```diff
@@ -1,5 +1,7 @@
-function reloadWeapon()
-	reloadPedWeapon(client)
+local function reloadWeapon()
+	if client then
+		reloadPedWeapon(client)
+	end
 end
 addEvent("relWep", true)
-addEventHandler("relWep", resourceRoot, reloadWeapon)
+addEventHandler("relWep", root, reloadWeapon)
```
Comment (jlillis) [inline [gameplay]/reload/reload.lua:7]: It would be better to leave this as `resourceRoot`, [according to the wiki](https://wiki.multitheftauto.com/wiki/TriggerServerEvent).

[gameplay]/reload/reload_c.lua
```diff
@@ -13,17 +13,20 @@ local blockedTasks = {
...
 }
 
-local function reloadWeapon()
+local function reloadTimer()
     if blockedTasks[getPedSimplestTask(localPlayer)] then
         return
     end
-    triggerServerEvent("relWep", resourceRoot)
+
+    triggerServerEvent("relWep", localPlayer)
```
Comment (jlillis) [inline [gameplay]/reload/reload_c.lua:21]: Again, better to stick with `resourceRoot` here.

jlillis (changes requested):

[gameplay]/reload/reload.lua
Comment (majqqq) [reply [gameplay]/reload/reload.lua:7]: I disagree.
theElement (sourceElement) is a 3rd argument in triggers. Bound element in event handlers defines which elements can actually trigger an event function, so it's a completely different thing. Some more information:
image
image

[gameplay]/reload/reload.lua
Comment (majqqq) [reply [gameplay]/reload/reload.lua:2]: If it's guaranteed i'm wondering why it's still appearing in this article, perhaps it is outdated a bit (last edit in 2017). I will push commit after your opinion on resourceRoot.

[gameplay]/reload/reload.lua
Comment (jlillis) [reply [gameplay]/reload/reload.lua:7]: Had to re-read this a few times - you're right, the disclaimer was for triggering events as root, not attaching event handlers to root.

[gameplay]/reload/reload.lua
Comment (jlillis) [reply [gameplay]/reload/reload.lua:2]: The advice, as I understand it, is that the variable `client` can be trusted to be the client that triggered the event. This is different from the `source` of the event - theoretically, a malicious client could trigger a server event and pass a different player as the source. For this resource the client is always going to be the source so we don't need to bother with checking anything.

Comment (majqqq) [reply [gameplay]/reload/reload.lua]: Done.

jlillis (approved):

PR #293, opened 2021-02-13
Fix webstats

patrikjuvonen:
## Summary
- Use Chart.js, remove swfs
- Increase update frequency to 5 seconds from 60 seconds
- Ability to expand and collapse resources in the right sidebar
- Add auto refresh checkbox that updates every 5 seconds
- Refactor code a little bit

Untitled1
Untitled2
Untitled3


jlillis (approved):
Tested and works well.

PR #295, opened 2021-03-02
Fixed bugs introduced by eef212e (admin: fix output message on actions)

Dutchman101:
PR #285 bugged name output

### DIFF
--- a/[admin]/admin/server/admin_server.lua
+++ b/[admin]/admin/server/admin_server.lua
@@ -521,28 +521,38 @@ end
 addCommandHandler ( "admin", aAdminMenu )
 
 function aAction ( type, action, admin, player, data, more )
-	if ( aLogMessages[type] ) then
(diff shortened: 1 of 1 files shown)


Dutchman101:
The reason this fixes it can be obscure, so let me just clarify

```lua
string = string.gsub ( string, "$player", aEscapeNickname( getPlayerName ( player ) ) )
```

In the docs the string.gsub function has the signature: string.gsub(s, pattern, replace [, n]), where n is the optional fourth argument which can be used to limit the number of substitutions made. When n is set to 0 then no substitutions are made - this is the problem, aEscapeNickName returns two values and just so it happens when the players nick does not contain "%" it sets n to 0 which means  no substitutions are made.

I will ask PR author to test their changes next time, I assumed they did and only looked at the code. I won't assume anything in the future.

PR #296, opened 2021-03-02
Add admin chat command

ricksterhd123:
Added a command handler which triggers aAdminChat and made the command name a setting which can be changed.


Dutchman101 (approved):

xLive:
I have some notes on this:


1. It doesn't use the correct permission `general.tab_adminchat` 
2. It should use `table.concat` instead because its faster and less code 
3. The command should be familiar command /adminchat? and it shouldn't be a setting? just like other commands
4. it should be added in admin2 too?

jlillis:
I agree with the above, and would add that feature additions to `admin` must be added to `admin2` as well.

Dutchman101:
@xLive 

3. The command should be really short like /a or it would partially defy the purpose: being a quick low effort way to write messages in admin chat. Besides, i know of some gamemodes that already use /a for other purposes, which is why the customisation doesn't hurt, it is easy compatibility out of the box.
4. That would be nice, but it's not a hard requirement

You're right about the rest, i was too fast. @ricksterhd123

PR #297, opened 2021-03-03
Addendum to #296

ricksterhd123:
I added 1 and 2 from this post:

> I have some notes on this:
> 
> 
> 1. It doesn't use the correct permission `general.tab_adminchat` 
> 2. It should use `table.concat` instead because its faster and less code 
> 3. The command should be familiar command /adminchat? and it shouldn't be a setting? just like other commands
> 4. It should be added in admin2 too?
> 
> _Originally posted by @xLive in https://github.com/multitheftauto/mtasa-resources/issues/296#issuecomment-789112953_


ricksterhd123:
I've added the same changes to admin2 as suggested. I'm not sure if the formatting is correct please let me know. I should've made it an issue first but I got carried away sorry for the screw up.

[admin]/admin/server/admin_server.lua
```diff
@@ -1570,12 +1570,8 @@ end )
 
 addCommandHandler(get("adminChatCommandName"), 
 	function(thePlayer, cmd, ...) 
-		if hasObjectPermissionTo (thePlayer, "general.adminpanel", false) then
-			local msg = ""
-			for i, word in ipairs(arg) do
-				msg = msg .. (i > 1 and " " or "") .. word
-			end
-			triggerEvent("aAdminChat", thePlayer, msg)
+		if hasObjectPermissionTo (thePlayer, "general.tab_adminchat", false) then
```
Comment (xLive) [inline [admin]/admin/server/admin_server.lua:1573]: Maybe a check if the message is not empty here? to avoid empty messages.

Comment (ricksterhd123) [reply [admin]/admin/server/admin_server.lua:1573]: We could use this check for adminchat gui as well

Comment (ricksterhd123) [reply [admin]/admin/server/admin_server.lua]: Nevermind I will leave the client-side alone as it's not really an issue. I've added those changes.

iDannz-Breno (approved):

jlillis (approved):

PR #299, opened 2021-03-04
editor: fix camera stuck when doubleclick on bottom icons

### LINKED ISSUE #298: Camera Stuck Mid-air after Browsing Models
**Describe the bug**

Hello, There's a bug with the Map Editor's Model Browser/Camera. sometimes when you Select an object with the Model Browser and you click on OK to place it, the Camera gets stuck Mid-air and doesn't move unless I disconnect from the Map Editor. I've reinstalled MTASA several times and tested with 1.5.7 and 1.5.8, but this still sometimes happens.

**To reproduce**

1. Open the Map Editor
2. Browse for any Object and place it
3. Keep placing until Camera gets stuck.
4. Camera got stuck and you can't move it until you Disconnect from the Map Editor.

**Expected behaviour**

Model Browser should place objects, not Making the Camera stuck.

**Screenshots**

Video: https://i.imgur.com/J9CLYea.mp4

**Version**

Tested on 1.5.7 and 1.5.8
Windows 7 SP1 (No Wine or anything)

**Additional context**

If there are any solutions, please tell me. Thank you

andr0xy:
This PR fixes #298 by adding a checker when doubleclick on the icons, placed on left-bottom screen.

Before: https://streamable.com/mu1xwe
Now: https://streamable.com/0t53o7

### DIFF
--- a/[editor]/editor_gui/client/buttonclicked.lua
+++ b/[editor]/editor_gui/client/buttonclicked.lua
@@ -31,27 +31,32 @@ end
 
 
 ---This is for when one of the element icons is clicked
+local clickTick = getTickCount()
 function elementIcons_Clicked ( source, mouseButton )
-	if mouseButton == "left" then
-		local elementType = elementIcons[source]["elementName"]
-		local resourceName = elementIcons[source]["resource"]
-		--triggerServerEvent ( "edfCreateElement", getRootElement(), elementType, resourceName )
-		local shortcut = resourceElementDefinitions[resourceName][elementType].shortcut -- E.g. immediately open model browser
-		call(getResourceFromName("editor_main"),"doCreateElement", elementType, resourceName, false, true, shortcut )
-	elseif mouseButton == "middle" then
-		local elementType = elementIcons[source]["elementName"]
-		local resourceName = elementIcons[source]["resource"]
-		--triggerServerEvent ( "edfCreateElement", getRootElement(), elementType, resourceName )
-		call(getResourceFromName("editor_main"),"doCreateElement", elementType, resourceName )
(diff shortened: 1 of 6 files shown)


Dutchman101:
Did this bug get introduced recently, and if so by what change?

Wirmaple73:
Thank you so much for the patched ButtonClicked.lua @andr0xy :+1: 
It worked, The Developers should replace the broken LUA with this one.

ricksterhd123:
> Did this bug get introduced recently, and if so by what change?
No, I checked blame and it's been unchanged for 12 years

patrikjuvonen:
Kind of a hacky fix. What if the call takes longer than a second to complete? Some sort of callback would be great instead.

Wirmaple73:
> 
> 
> Did this bug get introduced recently, and if so by what change?

I think it wasn't present in 1.5.6 because I never had problems with it, so this bug should have appeared since 1.5.7

andr0xy:
Thanks for the comments.

> Kind of a hacky fix. What if the call takes longer than a second to complete? Some sort of callback would be great instead.

Yes, you are right. I tested using fakelag command, with 5000+ ping, and the camera gets stuck again due to the bug. How could I implement a callback to this code? I spent like 2 hours trying to do it, but no success.

iDannz-Breno:
I don't think we need callbacks in this case, but just some state variables, to do not allow to toggle what is already toggled

Dutchman101:
> I don't think we need callbacks in this case, but just some state variables, to do not allow to toggle what is already toggled

This revision of the PR seems perfect to resolve both the bug and patrik's concern, thanks!

PR #305, opened 2021-03-19
mapcycler: misc. improvements

jlillis:
When merged, this PR will:
- Improve the settings definitions for display in the admin panel
- Add a new setting (hibernate_when_empty) to pause cycling while the server is empty. This saves resources that would otherwise be wasted on map changes on an empty server.


Dutchman101:
LGTM

PR #313, opened 2021-04-11
hedit: remove multiple blank lines

iDannz-Breno:


Dutchman101:
Thanks, implemented through 6b18053 instead (to ensure it matches what was applied to all other resources on the repository)

PR #316, opened 2021-04-11
admin: Improve lists

iDannz1:
After merged, this PR will add a filter to the following lists: 
-    Slap!
-    Give weapon
-    Give Vehicle

Before:
image

After:
image

_IMHO we don't need to add this to admin2, because their lists are a little bit different._


Dezash (commented):
The min version in meta.xml needs to be bumped
image

It would also make sense to apply this feature to Skins and Weathers in Server tab.

iDannz-Breno:
> It would also make sense to apply this feature to Skins and Weathers in Server tab.

I am not sure if we really should add this to Skins and Weathers because they are not like the lists mentioned in this PR, I can change it too, but I think it fits better in another PR.

Skins selection: 
image

Weather selection:
image

[admin]/admin/client/gui/admin_main.lua
```diff
@@ -122,25 +151,28 @@ y=y+B  aTab1.VehicleHealth	= guiCreateLabel ( 0.26, y, 0.25, 0.04, "Vehicle Heal
...
-								local row = guiGridListAddRow ( aTab1.VehicleOptions )
-								guiGridListSetItemText ( aTab1.VehicleOptions, row, 1, info.name, false, false )
-								guiGridListSetItemData ( aTab1.VehicleOptions, row, 1, tostring ( info.model ) )
-							end
+		
+		aTab1.GiveVehicle = guiCreateList( 0.71, 0.710, 0.27, 0.04, 0.275, "Give: "..getVehicleNameFromModel ( aCurrentVehicle ), true, aTab1.Tab, 'givevehicle')
+		
+		local vehicles = {}
+		for i = 400, 611 do
+			if (getVehicleNameFromModel(i) ~= "") then
```
Comment (Dezash) [inline [admin]/admin/client/gui/admin_main.lua:159]: The result of getVehicleNameFromModel could be stored in a variable so that you don't have to call it two times in a lengthy for loop

[admin]/admin/client/gui/admin_main.lua
```diff
@@ -49,13 +49,26 @@ function aAdminMenu ()
...
+
+		guiListSetColumns(aTab1.Slap, {{text = '', width = 0.8}})
+		guiListSetItems(aTab1.Slap, slaps)
+		guiListSetCallBack(aTab1.Slap, function(selectedData, selectedText)
+			local slap = tonumber(selectedData)
+			if slap then
+				aCurrentSlap = slap
+				guiSetText(aTab1.Slap, "Slap! "..slap..'  _')
+				if (aSpecSlap) then 
+					guiSetText(aSpecSlap, "Slap! "..slap.."hp")
```
Comment (Dezash) [inline [admin]/admin/client/gui/admin_main.lua:67]: It looks like `aSpecSlap` isn't defined anywhere so this is probably redundant

[admin]/admin/client/admin_gui.lua
```diff
@@ -51,3 +53,122 @@ function guiCreateButton ( x, y, w, h, text, relative, parent, right )
...
+		if (type(callback) == 'function') then
+			local row = guiGridListGetSelectedItem(gridlist)
+			local data = guiGridListGetItemData(gridlist, row, 1)
+			local text = guiGridListGetItemText(gridlist, row, 1)
+			if (row > -1) then
+				callback(data, text)
+				guiListSetVisible(list, false)
+			end
+		end
+	end)
```
Comment (Dezash) [inline [admin]/admin/client/admin_gui.lua:108]:
```suggestion
	end, false)

```

Dezash (commented):
To close the list, I now have to press a button when previously I could just click somewhere else. I think this behavior should be retained.

[admin]/admin/client/gui/admin_main.lua
Comment (iDannz-Breno) [reply [admin]/admin/client/gui/admin_main.lua:67]: Yes, I left that because it was already there before touching the code, it seems that this variable has been changed to `aSpectator.Slap`, I will replace it soon. Thanks.

[admin]/admin/client/gui/admin_main.lua
Comment (Pirulax) [reply [admin]/admin/client/gui/admin_main.lua:159]: The check is useless anyways. `getVehicleNameFromModel` never returns `""`

Comment (iDannz-Breno) [reply [admin]/admin/client/gui/admin_main.lua:159]: According to the wiki, it can return an empty string but I just tested it in-game and it looks like this <https://bugs.multitheftauto.com/view.php?id=8523> has already been fixed, I'll update the wiki page, thank you.

image

Comment (Pirulax) [reply [admin]/admin/client/gui/admin_main.lua:159]: Sorry, wrong wording. It does return an empty string, but only if the model isn't valid, which, in this case never happens.

Comment (iDannz-Breno) [reply [admin]/admin/client/gui/admin_main.lua:159]: image

Comment (Pirulax) [reply [admin]/admin/client/gui/admin_main.lua:159]: Sorry, I was looking at the C++ code, didn't look at the Lua API side.

Comment (iDannz-Breno) [reply [admin]/admin/client/gui/admin_main.lua:159]: No problem, thank you :D

Comment (iDannz-Breno) [reply [admin]/admin/client/gui/admin_main.lua:159]: btw wiki page updated too

iDannz-Breno:
I think it's ready to review again.

iDannz-Breno:
@jlillis, I can add such things for admin2, but afaik the lists are a little bit different in admin2. 
Btw If you want, feel free to merge, I think this weekend I can take a look at admin2 and submit another PR.

iDannz-Breno:
> @jlillis, I can add such things to admin2, but afaik the lists are a little bit different in admin2.
> Btw If you want, feel free to merge, I think this weekend I can take a look at admin2 and submit another PR.

resolved in #324.

jlillis:
Looks like there's a merge conflict blocking this - could you resolve?

iDannz1:
> Looks like there's a merge conflict blocking this - could you resolve?

@jlillis done

PR #317, opened 2021-04-16
admin: screenshot tweaks

theSarrum:
When merged, this PR will:
- Optimize bandwidth usage by using the _client_ global variable instead of passing the player by parameter.
- Fix permission checking by using an existing action name.

### DIFF
--- a/[admin]/admin/client/gui/admin_screenshot.lua
+++ b/[admin]/admin/client/gui/admin_screenshot.lua
@@ -16,10 +16,10 @@ function aPlayerScreenShot (player)
 		local x,y = guiGetScreenSize()
 		aScreenShotForm		= guiCreateWindow	( x / 2 - 300, y / 2 - 125, 600, 250, "Screenshot Management", false )
 		aScreenShotList		= guiCreateGridList	( 0.03, 0.08, 0.70, 0.90, true, aScreenShotForm )
-		aScreenShotNew		= guiCreateButton	( 0.75, 0.08, 0.42, 0.09, "Take New", true, aScreenShotForm )
-		aScreenShotDelete	= guiCreateButton	( 0.75, 0.18, 0.42, 0.09, "Delete", true, aScreenShotForm )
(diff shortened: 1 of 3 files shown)


Dezash:
I think it would be better to add the `takescreenshot` permission to the default ACL configuration rather than remove it

theSarrum:
> I think it would be better to add the `takescreenshot` permission to the default ACL configuration rather than remove it

I thought about it, but decided that it makes no sense to restrict screenshots for certain ACL groups, because they do not contain any private information.
Then there are 2 options:
1. Add one permission (`command.screenshots`?) that will block full access to screenshots
2. Add different permissions for different actions:
2.1. 'Take New' - `command.takescreenshot`
2.2. 'Delete' - `command.deletescreenshot`
2.3. 'View' - `command.viewscreenshot`
2.4. 'Refresh' - `command.listscreenshots`

Dezash:
More granular permissions are better in my opinion. You may not want trial moderators deleting screenshots or sometimes even taking them. However, the mentioned permissions should use "general" prefix instead of "command".

theSarrum:
I agree with the prefix, but if we want to block buttons, there is no other option.
https://github.com/multitheftauto/mtasa-resources/blob/44053e2c7b59b84183cf1a349822c853354aa76f/%5Badmin%5D/admin/client/admin_gui.lua#L43

Pirulax:
I think you should be able to change the triggers on client side to be `localPlayer` instead of `resoureRoot`.

theSarrum:
I have made changes, now we can restrict the rights for certain groups.

> I think you should be able to change the trigger sources on client side to be `localPlayer` instead of `resoureRoot`, then just attach the handler to root on the server side.

Not sure if this is necessary. But if others agree, I'll do it.
Also weird that then we will have to attach to `root` on the server instead of `resourceRoot`.

srslyyyy:
> Not sure if this is necessary. But if others agree, I'll do it.
Why not. Would be nice to follow this in any upcoming PR, if possible.

theSarrum:
> Why not. Would be nice to follow this in any upcoming PR, if possible.

Done.

jlillis (approved):

PR #320, opened 2021-04-25
Fix some typo in interiors.xml

Allerek:
According to fandom, its Didier Sachs.
https://gta.fandom.com/wiki/Didier_Sachs

image


Allerek:
Reopened for merging.

PR #315, opened 2021-04-11
admin: some improvements 

iDannz1:
When merged, this PR will:

- Merge the PositionX, PositionY, PositionZ labels into one.

- Move and re-scale "unread messages" and "screenshots" buttons from the "Player" tab.

- Add new functionality: Hide Sensitive Data.
    - Useful for those who stream, share screens, or just want to share a simple screenshot. Works like "Streamer mode" in some apps, but in this case, it will never allow the admin panel to display the following data:
        - Players Tab:
            - Player serial
            - Player IP
            - Player account name
            - Player in-game localization
        - Server Tab:
            - Server password
                 
- Load and save the following data in client-side XML:
     - Anonymous Admin status
     - Hide Color Codes status
     - Hide Sensitive Data status
     
_If it's approved, ill do the same for admin2._

Before:
image

After:
image

PR #325, opened 2021-05-24
admin2: add "hide sensitive data"

iDannz1:
As requested in #315, i've added these changes to admin2. (check #315's conversation for more details)


jlillis:
For the `admin2` implementation, I think this checkbox should be moved to the "Options" tab. There are a other minor UI improvements I have in mind, but that is the only one that pertains to this PR.

jlillis:
Ignore my previous comment - I'll address this in a separate issue. There are other panel options that are missing from the options tab.

PR #326, opened 2021-05-29
Update admin_servermaps.lua

1STD:
I made utils functions to return a boolean, this function is checking whether the feature is active or not, so it will return a false or a true one.
Why did you do that?
Well, when you go to the "Maps" tab, a variable will return an error, if the resource is not linked to the variable and leading to null within ipairs.

### DIFF
--- a/[admin]/admin/server/admin_servermaps.lua
+++ b/[admin]/admin/server/admin_servermaps.lua
@@ -160,15 +160,3 @@ function sortCompareFunction(s1, s2)
         return false
     end
 end
-
-function stateResourceName(resourceName, state)
-	if resourceName then
-		local resourceFromName = getResourceFromName(resourceName);
-
-		if getResourceState(resourceFromName) == tostring(state) then
-			return true;
-		else
-			return false;
-		end
-	end
-end
--- a/[admin]/admin/server/admin_servermaps.lua
+++ b/[admin]/admin/server/admin_servermaps.lua
@@ -10,7 +10,10 @@
 
 function getServerMaps (loadList)
 	if checkClient( true, source, 'getServerMaps' ) then return end
-	local tableOut
+	local tableOut = {}
+
+	-- It will check if the feature is enabled.
+	if not stateResourceName("mapmanager", "running") then return end
 
 	-- Check if 'mapmanager' resource exists
 	if ( not getResourceFromName("mapmanager") ) then
@@ -20,7 +23,6 @@ function getServerMaps (loadList)
 	end
 
 	if loadList then
-		tableOut = {}
 		-- local deletedMaps = {}
 		local gamemodes = {}
 		gamemodes = call(getResourceFromName("mapmanager"), "getGamemodes")
@@ -157,4 +159,16 @@ function sortCompareFunction(s1, s2)
     else
         return false
     end
+end
+
+function stateResourceName(resourceName, state)
+	if resourceName then
+		local resourceFromName = getResourceFromName(resourceName);
+
+		if getResourceState(resourceFromName) == tostring(state) then
+			return true;
+		else
+			return false;
+		end
+	end
 end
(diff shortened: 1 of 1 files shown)


iDannz1:
Well, thank you for the pull request, the idea is very good and I also have this same problem of ending up stopping the resource `mapmanager` and having this annoying error on the server console, but I think the implementation you applied could be much simpler and more advantageous, without having to add any new functions and also preventing multiple calls to `getResourceFromName` in the function `getServerMaps`.

And another thing, within the original code, [there is already a check if the resource exists](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/admin/server/admin_servermaps.lua#L16), and if it does not exist it sends a trigger to the client with an empty table as result, and with a comment of "fake data". This behavior shouldn't be expected if the resource exists but is not being executed? (I didn't dive into the original code to find out if the result is really expected to be achieved - even if it is "fake")

[admin]/admin/server/admin_servermaps.lua
```diff
@@ -10,7 +10,10 @@
...
 function getServerMaps (loadList)
 	if checkClient( true, source, 'getServerMaps' ) then return end
-	local tableOut
+	local tableOut = {}
+
+	-- It will check if the feature is enabled.
+	if not stateResourceName("mapmanager", "running") then return end
 
 	-- Check if 'mapmanager' resource exists
 	if ( not getResourceFromName("mapmanager") ) then
```
Comment (iDannz1) [inline [admin]/admin/server/admin_servermaps.lua:19]:
```suggestion
	if ( not mapmanager ) then
```

[admin]/admin/server/admin_servermaps.lua
```diff
@@ -158,3 +160,15 @@ function sortCompareFunction(s1, s2)
...
+	if resourceName then
+		local resourceFromName = getResourceFromName(resourceName);
+
+		if getResourceState(resourceFromName) == tostring(state) then
+			return true;
+		else
+			return false;
+		end
+	end
+end
```
Comment (iDannz1) [inline [admin]/admin/server/admin_servermaps.lua:174]:
```suggestion
```

[admin]/admin/server/admin_servermaps.lua
```diff
@@ -20,7 +23,6 @@ function getServerMaps (loadList)
 	end
 
 	if loadList then
-		tableOut = {}
 		-- local deletedMaps = {}
 		local gamemodes = {}
 		gamemodes = call(getResourceFromName("mapmanager"), "getGamemodes")
```
Comment (iDannz1) [inline [admin]/admin/server/admin_servermaps.lua:28]:
```suggestion
		gamemodes = call(mapmanager, "getGamemodes")
```

[admin]/admin/server/admin_servermaps.lua
```diff
@@ -10,7 +10,10 @@
 
 function getServerMaps (loadList)
 	if checkClient( true, source, 'getServerMaps' ) then return end
-	local tableOut
+	local tableOut = {}
+
+	-- It will check if the feature is enabled.
+	if not stateResourceName("mapmanager", "running") then return end
```
Comment (iDannz1) [inline [admin]/admin/server/admin_servermaps.lua:16]:
```suggestion
	local mapmanager = getResourceFromName("mapmanager")
	
	if mapmanager and (getResourceState(mapmanager) ~= "running") then
	    mapmanager = nil
	end
```

iDannz1 (commented):
there are still a few other places (lines: 34, 43, 57 and 58) to replace `getResourceFromName("mapmanager")` to the `mapmanager` variable, but for some reason, i cant select them to comment on)

iDannz1:
Now it looks good, maybe we should check if the same bug exists in admin2?

1STD:
> Now it looks good, maybe we should check if the same bug exists in admin2?

I'll just wait.

iDannz1 (approved):
I just checked and the admin2 resource doesn't use `mapmanager` anywhere.

PR #329, opened 2021-06-09
admin: replace "give admin" button with permissions widget

iDannz1:
I will send another PR to do the same/similar to admin2 in near future.

image


xLive:
I've an idea for the design, something like this:
image

What do you think?

iDannz1:
Looks good and I can apply something similar, but first lets wait for other opinions.

Allerek:
> I've an idea for the design, something like this:
> image
> 
> What do you think?
Thats cool, simple and easy to understand, even newbies will understand that design.

Dutchman101:
> I've an idea for the design, something like this:
> image
> 
> What do you think?

I like this design more than the current one in PR, just add something on the left side to clarify those are the _available_ groups (to assign) and something on the right to clarify those are the _current_ groups for that user

iDannz1:
bruhhh looks like i just ruined my branch

iDannz1:
I think I finally managed to fix it. Sorry for the useless commits, let me know if you think is better to create a new PR, I still learning this thing called git. 

Anyways, this is the new design:
image
_Please comment below if you have any suggestions._

jlillis (approved):

Dutchman101 (approved):

PR #334, opened 2021-07-02
speedometer: Check vehicle occupied by localPlayer

ricksterhd123:
When onClientElementModelChange is triggered it doesn't check that localPlayer is an occupant of source. So when you're inside a plane it sometimes showsSpeedo randomly (when the event is triggered elsewhere)

### DIFF
--- a/[gameplay]/speedometer/client.lua
+++ b/[gameplay]/speedometer/client.lua
@@ -62,6 +62,7 @@ addEventHandler("onClientElementDestroy", root, destroyHandler)
 -- If player vehicle changes in abnormal way (e.g drives into a vehicle pick-up)
 local function onVehicleTypeChange(oldModel, newModel)
     if (getElementType(source) ~= "vehicle") then return end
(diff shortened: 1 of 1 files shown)


[gameplay]/speedometer/client.lua
```diff
@@ -62,6 +62,7 @@ addEventHandler("onClientElementDestroy", root, destroyHandler)
 -- If player vehicle changes in abnormal way (e.g drives into a vehicle pick-up)
 local function onVehicleTypeChange(oldModel, newModel)
     if (getElementType(source) ~= "vehicle") then return end
+    if (getPedOccupiedVehicle(localPlayer) ~= source) then return end
```
Comment (xLive) [inline [gameplay]/speedometer/client.lua:65]:
```suggestion
    if (getPedOccupiedVehicle(localPlayer) ~= source) then return end
```
I don't think `getElementType` check is necessary anymore

iDannz1 (approved):

PR #337, opened 2021-08-04
Bump admin version number for upcoming release

jlillis:
This PR bumps `admin`s version number for the next release. As recommended by @iDannz1 in Discord, this will help advertise the fact that admin has had a number of improvements since the last release cycle and hopefully encourage admins to update.

### DIFF
--- a/[admin]/admin/client/admin_client.lua
+++ b/[admin]/admin/client/admin_client.lua
@@ -10,7 +10,7 @@
 
 _DEBUG = false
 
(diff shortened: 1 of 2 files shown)


iDannz1 (changes requested):
Looks like you forgot to change here too
https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/admin/admin_definitions.lua#L13

PR #339, opened 2021-08-16
Push improvements.

srslyyyy:
- Replaced loops.
- Replaced event sources.
- Localized LOD_Map table (it is not used anywhere else).
- Replaced semicolons with commas.
- Improved readability.

I did small test, but would be nice if someone could test it accurately.


iDannz1:
Hey, thanks for this PR :D

I think we must keep the event attachments in `resourceRoot`, or the events will be triggered by other map resources...

Imagine this scenario: you have `10 resource maps` (`map1`, `map2`, ..., `map10`). When the resource `map1` starts, it will call the server-side event `requestLODsClient` of each map resource below the `root` element (this case: `map2`, `map3`, ..., `map10`), and they will return the content of the tables `usedLODModels` to all the other maps via the client-side event `setLODsClient`, and all they will all apply `engineSetModelLODDistance`, whereas in this case we just need to sync it once and within the same resource.

srslyyyy:
@iDannz1 yeah good point, i forgot that people could use multiple map resources, i'm used to small resources count.

jlillis (approved):
Looks good and appears to work, although I don't have many maps to test with. Perhaps someone with more editor/mapping experience could check this out?

[editor]/editor_main/server/mapEditorScriptingExtension_s.lua
```diff
@@ -3,43 +3,65 @@
...
-				local rx,ry,rz = getElementRotation(object)
-				local lodObj = createObject(lodModel,x,y,z,rx,ry,rz,true)
-				setElementInterior(lodObj, getElementInterior(object) )
-				setElementDimension(lodObj, getElementDimension(object) )
-				setElementParent(lodObj, object)
-				setLowLODElement(object, lodObj)
-				usedLODModels[lodModel] = true
+
+	if startEvent then
+		local thisResource = getThisResource()
```
Comment (Zangomangu) [inline [editor]/editor_main/server/mapEditorScriptingExtension_s.lua:32]: Can use the resource parameter which is passed from the event instead since its bound to resourceRoot

Comment (srslyyyy) [reply [editor]/editor_main/server/mapEditorScriptingExtension_s.lua]: Done.

Zangomangu (approved):

PR #340, opened 2021-08-18
editor: Added collision patches for interiors

Zangomangu:
@tut98 has modelled collision for further 853 objects, mainly inside interiors, which are now selectable in the editor.


Dutchman101 (approved):
Looks good to me, but due to its size we'll have to rely on users testing it in practise as well.

PR #342, opened 2021-08-19
killmessages: fix dup rows

iDannz1:
This PR fixes the bug that duplicates the displayed lines but also ends up disabling ``freeroam`` suicide detection, so we need another method to detect it. ~~Anyway, I'm going to open an issue requesting this functionality back.~~


Dutchman101:
Basically was a workaround for the (just now filed) issue #2312 .. and it didn't turn out well. Thanks for the quick fix!

PR #347, opened 2021-08-24
admin: some tweaks

iDannz1:
This PR replaces deprecated functions, removes unused code, fixes calls to non-existing functions, etc...

I found them using [this](https://github.com/sumneko/vscode-lua), and maybe in the near future, i can send other PRs doing the same to other resources.

### DIFF
--- a/[admin]/admin/client/gui/admin_vehicle.lua
+++ b/[admin]/admin/client/gui/admin_vehicle.lua
@@ -241,9 +241,9 @@ function aClientVehicleClick ( button )
 			triggerServerEvent ( "aVehicle", getLocalPlayer(), aVehicleCustomizePlayer, "setlights", { guiGetText ( aLightsColor ) } )
 		elseif ( source == aLicensePlateSet ) then
 			triggerServerEvent ( "aVehicle", getLocalPlayer(), aVehicleCustomizePlayer, "setplates", { guiGetText ( aLicensePlate ) } )
-		elseif ( source == aVehicleColorScheme ) then
-			guiSetVisible ( aVehicleColorForm, true )
-			guiBringToFront ( aVehicleColorForm )
+		-- elseif ( source == aVehicleColorScheme ) then
+		-- 	guiSetVisible ( aVehicleColorForm, true )
+		-- 	guiBringToFront ( aVehicleColorForm )
 		elseif ( source == aVehicleUpgradeAll ) then
 			triggerServerEvent ( "aVehicle", getLocalPlayer(), aVehicleCustomizePlayer, "customize", { "all" } )
 			setTimer ( aVehicleCheckCurrentUpgrades, 2000, 1, aVehicleCustomizeVehicle )
--- a/[admin]/admin/server/admin_sync.lua
+++ b/[admin]/admin/server/admin_sync.lua
@@ -23,7 +23,6 @@ function aSynchCoroutineFunc( type, data, typeOfTag, banSearchTag )
 		tableOut["mute"] = isPlayerMuted ( data )
 		tableOut["freeze"] = isPlayerFrozen ( data )
 		tableOut["money"] = getPlayerMoney ( data )
-		tableOut["username"] = getPlayerUserName ( data ) or "N/A"
 		tableOut["version"] = aPlayers[data]["version"]
 		tableOut["accountname"] = getPlayerAccountName ( data ) or "N/A"
 		tableOut["groups"] = "None"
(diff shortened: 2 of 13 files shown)


[admin]/admin/client/gui/admin_vehicle.lua
```diff
@@ -241,9 +241,9 @@ function aClientVehicleClick ( button )
 			triggerServerEvent ( "aVehicle", getLocalPlayer(), aVehicleCustomizePlayer, "setlights", { guiGetText ( aLightsColor ) } )
 		elseif ( source == aLicensePlateSet ) then
 			triggerServerEvent ( "aVehicle", getLocalPlayer(), aVehicleCustomizePlayer, "setplates", { guiGetText ( aLicensePlate ) } )
-		elseif ( source == aVehicleColorScheme ) then
-			guiSetVisible ( aVehicleColorForm, true )
-			guiBringToFront ( aVehicleColorForm )
```
Comment (iDannz1) [inline [admin]/admin/client/gui/admin_vehicle.lua:246]: this GUI element was disabled some time ago
https://github.com/multitheftauto/mtasa-resources/blob/44053e2c7b59b84183cf1a349822c853354aa76f/%5Badmin%5D/admin/client/gui/admin_vehicle.lua#L77

[admin]/admin/server/admin_sync.lua
```diff
@@ -122,20 +122,6 @@ function aSynchCoroutineFunc( type, data, typeOfTag, banSearchTag )
 		tableOut["map"] = getMapName()
 		tableOut["password"] = getServerPassword()
 		tableOut["fps"] = getFPSLimit()
-	elseif ( type == "rights" ) then
```
Comment (iDannz1) [inline [admin]/admin/server/admin_sync.lua:127]: This part is never executed, and even if it was, the code inside it doesn't make sense.

[admin]/admin/server/admin_sync.lua
```diff
@@ -122,20 +122,6 @@ function aSynchCoroutineFunc( type, data, typeOfTag, banSearchTag )
 		tableOut["map"] = getMapName()
 		tableOut["password"] = getServerPassword()
 		tableOut["fps"] = getFPSLimit()
-	elseif ( type == "rights" ) then
-		for gi, group in ipairs ( aclListGroups() ) do
```
Comment (iDannz1) [inline [admin]/admin/server/admin_sync.lua:128]: `aclListGroups` is not defined anywhere and it's not a default function

[admin]/admin/server/admin_sync.lua
```diff
@@ -122,20 +122,6 @@ function aSynchCoroutineFunc( type, data, typeOfTag, banSearchTag )
 		tableOut["map"] = getMapName()
 		tableOut["password"] = getServerPassword()
 		tableOut["fps"] = getFPSLimit()
-	elseif ( type == "rights" ) then
-		for gi, group in ipairs ( aclListGroups() ) do
-			for oi, object in ipairs ( aclGroupListObjects ( group ) ) do
-				if ( ( object == data ) or ( object == "user.*" ) ) then
-					for ai, acl in ipairs ( aclGroupListACL ( group ) ) do
-						for ri, right in ipairs ( aclListRights ( acl ) ) do
-							local access = aclGetRight ( acl, string )
```
Comment (iDannz1) [inline [admin]/admin/server/admin_sync.lua:133]: `string` is not defined anywhere

[admin]/admin/server/admin_sync.lua
```diff
@@ -122,20 +122,6 @@ function aSynchCoroutineFunc( type, data, typeOfTag, banSearchTag )
...
 		tableOut["password"] = getServerPassword()
 		tableOut["fps"] = getFPSLimit()
-	elseif ( type == "rights" ) then
-		for gi, group in ipairs ( aclListGroups() ) do
-			for oi, object in ipairs ( aclGroupListObjects ( group ) ) do
-				if ( ( object == data ) or ( object == "user.*" ) ) then
-					for ai, acl in ipairs ( aclGroupListACL ( group ) ) do
-						for ri, right in ipairs ( aclListRights ( acl ) ) do
-							local access = aclGetRight ( acl, string )
-							if ( access ) then table.insert ( tableOut, right ) end
```
Comment (iDannz1) [inline [admin]/admin/server/admin_sync.lua:134]: never executed, because `access` will always be false

jlillis (approved):
This looks good after a brief review. Nice work.

ricksterhd123:
Did you have to setup vscode-lua in some way to stop warning about mta functions? I remember trying similar tools and it wouldn't stop complaining about mta functions. I have a lot of old code to refactor and this tool would be great. Perhaps you could fork it and add MTA specific features such as spotting deprecated functions etc? I can imagine it would be very popular among scripters.

ricksterhd123:
On the same topic: Dutch showed me this before: https://github.com/Koihik/LuaFormatter and it looks promising perhaps we could configure a standard formating for mtasa resources. Only issues I had with it was that lambda functions would look weird but i think it's configurable. Shall we open a discussion? Maybe it's unecessary but I think it would be worth it.

iDannz1:
> On the same topic: Dutch showed me this before: https://github.com/Koihik/LuaFormatter and it looks promising perhaps we could configure a standard formating for mtasa resources. Only issues I had with it was that lambda functions would look weird but i think it's configurable. Shall we open a discussion? Maybe it's unecessary but I think it would be worth it.

Maybe we could use <https://github.com/multitheftauto/mtasa-resources/issues/250> to discuss this subject

Dutchman101:
Tested on a high traffic server for over a month, and confirmed to be stable.

PR #352, opened 2021-08-27
Optimization of gamemode play

1lyncon:
I know that a lot of people don't use the "play" gamemode much, as I don't use it anymore I decide to make it much more optimized than the previous one.

### DIFF
--- a/[gamemodes]/[play]/play/broph.lua
+++ b/[gamemodes]/[play]/play/broph.lua
@@ -26,8 +26,8 @@ addEventHandler("onResourceStart", resourceRoot, function()
 		spawn(players)
 	end
 
-	for k, v in pairs(vehsToSpawn) do
-		createVeh(v)
+	for index, value in ipairs(vehsToSpawn) do
+		createVeh(value)
 	end
 end)
 
--- a/[gamemodes]/[play]/play/broph.lua
+++ b/[gamemodes]/[play]/play/broph.lua
@@ -1,176 +1,132 @@
-local vehiclesToSpawn =
-{
-	{557, -676.43225097656, 962.98449707031, 12.507791519165, 92.063659667969},
-	{557, -676.67596435547, 968.88238525391, 12.507817268372, 89.458343505859},
-	{429, -696.71008300781, 947.44274902344, 11.961327552795, 43.260345458984},
-	{429, -700.66668701172, 946.43811035156, 12.032683372498, 41.218505859375},
-	{541, -708.88635253906, 939.77947998047, 12.079864501953, 180.60076904297},
-	{481, -679.98175048828, 931.08044433594, 11.648596763611, 88.025085449219},
-	{481, -679.97430419922, 932.28381347656, 11.64875125885, 87.565948486328},
-	{469, -655.177734375, 963.35192871094, 12.175182342529, 89.811737060547},
-	{468, -681.74114990234, 912.31903076172, 11.776277542114, 189.92042541504},
-	{468, -680.06512451172, 912.44622802734, 11.779381752014, 191.75022888184},
-	{468, -678.29089355469, 912.71087646484, 11.77979183197, 200.86720275879},
-	{493, -649.33892822266, 867.71429443359, -0.10552111268044, 228.00956726074},
-	{473, -661.28826904297, 876.15618896484, -0.30112856626511, 227.04469299316},
-	{484, -647.07635498047, 876.42327880859, 0.22773677110672, 226.89099121094},
-	{568, -696.35430908203, 929.35675048828, 12.135828971863, 178.99200439453},
-	{511, -696.59759521484, 897.68084716797, 13.838381767273, 176.86285400391},
-	{402, -686.66400146484, 969.24591064453, 11.975840568542, 89.380676269531},
-	{475, -687.01416015625, 963.00555419922, 11.94958114624, 89.711608886719},
-	{457, -686.92352294922, 966.03717041016, 11.767498970032, 89.163909912109},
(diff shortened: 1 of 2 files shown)


1lyncon:
> @ 1lyncon Em geral, pare de criar PRs para fechá-los logo em seguida. Se for por acidente, por favor, tenha mais cuidado ..

I'm sorry, it's because I was having a hard time with some things here, that's why I'm creating and recreating pr.

1lyncon:
> With these changes you did eliminate the settings? Whats the porpouse of this? You could have changed it in other way, like for example saving in a variable the new value of the setting every time it changed, so it doesn't use get() over and over again, but you now made it hardcoded, which is not as good.

Well, I removed the metasettings and it might make things easier that you don't need to pull with get(), it might make things in gamemode "play" easier, I don't know,

jlillis:
I'm sorry if this comes off as rude, but I don't see any optimization here - just deleted functionality. If you "don't know" then perhaps it's best to mark this PR as WIP for now and discuss what you're trying to do in the dev discord.

iDannz1:
First of all, thank you so much for your effort in trying to refactor gamemode play 😉.

I really agree that currently the default code of this gamemode is a bit weird and doesnt follow the guidelines that have been followed in other default resources (ATM we havent come to an agreement on this subject), but anyway i have a few points to cite about your PR:

1 - You removed most of the vehicle and player spawn points.

2 - You removed the MTA settings system implementation, which is used with the `get` function. This may sound positive to you, but trust me, in this case, it is not good. 

Using `get`, server owners can change these settings in real-time using other resources, such as the admin panel.
With your change, you have made them hard-coded, so now the server owners have to edit these values ​​manually within the script and they need to restart the resource for this change to take effect.

3 - There are some scripting errors in your script, i will show you in a few minutes.

And as jlillis said, I cant see any optimization here either 🙃

1lyncon:
Thank you very much for your answers, now that I understand perfectly, I'm not going to create pr for useless things like this I created, so I'll think well for myself to create the pr that the mtasa-resource needs. I'm closing to not involve problems or something like that

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,176 +1,132 @@
...
+local vehTimeToDestroy = {}
+local playerVeh = {}
 
-	if player and isElement(player) then
-		local x,y,z,r = unpack(spawnpoints[math.random(1,#spawnpoints)])
-		spawnPlayer(player,x+math.random(-3,3),y+math.random(-3,3),z,r,validSkins[math.random(1,#validSkins)],0,0)
-		fadeCamera(player, true)
-		setCameraTarget(player, player)
+addEventHandler("onResourceStop", resourceRoot, function()
+	destroyElements()
```
Comment (iDannz1) [inline [gamemodes]/[play]/play/broph.lua:21]: IMO this is not necessary, because when a resource is stopped, all its elements are destroyed by default. And on the other hand, this call also doesn't make sense, since the function ``destroyElements`` will interact with `source`, expecting it to be a player, and in this case, it isn't.

> onResourceStop: The source of this event is the root element of the resource that is being stopped.

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,176 +1,132 @@
...
+end)
+
+addEventHandler("onResourceStart", resourceRoot, function()
+	for _, players in ipairs(getElementsByType("player")) do
+		spawn(players)
```
Comment (iDannz1) [inline [gamemodes]/[play]/play/broph.lua:26]:
```suggestion
	for _, player in ipairs(getElementsByType("player")) do
		spawn(player)
```

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,176 +1,132 @@
...
 
-end
+addEventHandler("onPlayerQuit", root, function()
+	destroyElements()
+end)
 
-local function onWasted()
+addEventHandler("onPlayerWasted", root, function()
+	local playerRespawn = defaultTimes["PlayerRespawn"]
+	setTimer(spawn, (playerRespawn > 50 and playerRespawn or 50), 1, source)
```
Comment (iDannz1) [inline [gamemodes]/[play]/play/broph.lua:44]: Another tip: in these cases, you can simply use math.min or max.max to do the trick...
```suggestion
	setTimer(spawn, math.max(playerRespawn, 50), 1, source)
```
http://lua-users.org/wiki/MathLibraryTutorial

[gamemodes]/[play]/play/broph.lua
```diff
@@ -1,176 +1,132 @@
...
-	vehicle.damageProof = true
-	vehicle.frozen = true
-	vehiclesToSpawn[vehicle] = vehicledata
-	addEventHandler("onVehicleEnter",vehicle,onEnter)
-	addEventHandler("onVehicleExit",vehicle,onExit)
+	local vehCreated = createVehicle(id, x, y, z, 0, 0, r)
+	setVehicleDamageProof(vehCreated, true)
+	setElementFrozen(vehCreated, true)
 
+	vehsToSpawn[vehCreated] = data
```
Comment (iDannz1) [inline [gamemodes]/[play]/play/broph.lua:114]: It looks like this value is being added to the table, but it is never removed when the vehicle is destroyed...

iDannz1 (commented):
Just a quick review, I am not sure if this should be merged, but here is my feedback for you, good luck 👍

_For a better reading, I suggest you use the [split-screen](https://github.com/multitheftauto/mtasa-resources/pull/352/files?diff=split&w=1) mode and pay attention only to your additions._

iDannz1:
Also, taking into consideration the other pull requests you have opened, it looks like you are having problems with git, if that is the case, feel free to contact me via discord, I will be happy to help or refer you to some tutorials.

PR #354, opened 2021-09-24
Lua linter implementation

Disinterpreter:
Just fast Q&A to my PR

What is it? - https://en.wikipedia.org/wiki/Lint_(software)
Why is it need for **us**? - This repo is a library of examples for newbies and other scripters I think it must be pure and clear for all, even for our GC.
Is it need to merge now? - Not necessary. But if you want to see red crossmark on each commit about fixing the problems, you can. Anyway, you can set it as a "Draft".
How it looks? - https://github.com/Disinterpreter/mtasa-resources/runs/3697351842
Why are you ignore lot of marks in config? - Linter right now uses for common lua, and it hasn't our functions/variables/tables. We can fork it and change later.

Linter: https://github.com/mpeterv/luacheck

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,14 @@
+name: MTA Resources Linter
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


ciber96:
Nice

Dutchman101:
Let's trial it for a month and see how this works out in practise (for PR's & MTA's Lua stack)..

PR #357, opened 2021-10-16
editor_main: improve gamemodestopper

srslyyyy:
- Replaced bound element to predefined resourceRoot
- Used startedResource from parameters instead of method call getThisResource()
- Loop isn't called anymore when resource can't stop other resources (missing perms)
- Also it's no longer in it's scope (perm check)
- Replaced ipairs with better equivalent
- Improved readability

### DIFF
--- a/[editor]/editor_main/gamemodestopper.lua
+++ b/[editor]/editor_main/gamemodestopper.lua
@@ -1,5 +1,5 @@
 function isResourceRunning(res)
-	return getResourceState(res)=="running"
+	return getResourceState(res) == "running"
 end
 
 function isGamemode(res)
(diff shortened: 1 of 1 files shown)


[editor]/editor_main/gamemodestopper.lua
```diff
@@ -10,17 +10,28 @@ function isMap(res)
...
-				if hasObjectPermissionTo(getThisResource(), "function.stopResource") then
-					stopResource(resource)
-				else
-					outputDebugString("Editor: Unable to stop running gamemodes (no access to function.stopResource)")
-					return
-				end
+function onResourceStart(startedResource)
+	local stopPermission = hasObjectPermissionTo(startedResource, "function.stopResource")
+
+	if not stopPermission then
```
Comment (jlillis) [inline [editor]/editor_main/gamemodestopper.lua:16]: This could be shortened to `if not hasObjectPermissionTo(startedResource, "function.stopResource") then` rather than allocating a variable only used once. At least, that would be my style preference. It's not a blocker in my opinion.

[editor]/editor_main/gamemodestopper.lua
```diff
@@ -10,17 +10,28 @@ function isMap(res)
...
+		return false
+	end
+
+	local resourcesTable = getResources()
+
+	for resourceID = 1, #resourcesTable do
+		local resourceElement = resourcesTable[resourceID]
+		local resourceRunning = isResourceRunning(resourceElement)
+
+		if resourceRunning then
```
Comment (jlillis) [inline [editor]/editor_main/gamemodestopper.lua:28]: Same as above - `if isResourceRunning(resourceElement) then`

jlillis (approved):
Requested changed minor and optional - I think this is good to merge in a few days as-is if nobody else has blocking comments.

srslyyyy:
@jlillis done.

PR #364, opened 2021-11-02
enhancement: joinquit settings support + code refactor

itslewiswatson:
This PR introduces some quality of life improvements to the `joinquit` resource, as well as some basic enhancements:
- Moved the script server-side
- Added `*showColorCodes` setting which accepts a boolean value
- Added `*defaultColor` setting which accepts an RGB input, and controls what color to output the message text in
- Removed anonymous functions 
- Removed redundant support for alpha values which weren't present in hex code so I'm not sure why we used them
- Resource listens for settings changes
- Moved a few pieces of functionality into functions, namely `getDefaultColor` (which has a fallback now) and `getHexFriendlyNick` which makes the `outputChatBox` calls read nicer

### DIFF
--- a/[gameplay]/joinquit/joinquit.lua
+++ b/[gameplay]/joinquit/joinquit.lua
@@ -1,59 +1,65 @@
-local showColorCodes = true; 		-- Shows player's names colorcoded if set to true, and if set to false it doesn't
-local defaultHexCode = "#4E5768"; 	-- Hex code for what color to output messages in (only used if showColorCodes is true)
+local resourceName = (Resource.getThis()).name
+local settingPrefix = ("*%s."):format(resourceName)
 
+local showColorCodes = get("showColorCodes") == "true" 	-- Shows player"s names colorcoded if set to true, and if set to false it doesn"t
+local defaultColor = get("defaultColor")				-- Hex code for what color to output messages in (only used if showColorCodes is true)
+local fallbackHexCode = "#4E5768"						-- Fallback hex code for incorrectly input settings values
+
+function reloadSettings(settingName)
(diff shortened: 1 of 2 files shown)


srslyyyy:
> Refactored to use OOP

Don't get me wrong but OOP is mistake in MTA, i would avoid it at any cost, unless you can't do some code without it.
It's horrible in performance, and simply enabling it in meta would cause increased memory/cpu usage.
obraz

itslewiswatson:
Regardless of the perceived efficiency, it is simply much easier for most people to read and write object-oriented code. Which is frankly more important than micro-efficiencies and premature optimizations, especially for a project of this magnitude. Writing it this way means it is more accessible to more people, and much easier for people like me who context-switch a lot.

I doubt the overhead of OOP degrades performance by such a margin that could be reasonably referred to as "horrible". If you have serious concerns about my code using OOP, please raise them here. Otherwise I fear this is off-topic, but feel free to open an issue and we could discuss OOP usage in this repo potentially.

srslyyyy:
**most people to read and write object-oriented code**
Most of people uses POP, i can tell you that by observing scripting help channels from many years.
Anyways it's up to you, i just suggested refraining by using it :)

Dutchman101:
@itslewiswatson There is no clear coding guidelines yet, but we won't accept this PR if it uses OOP.. performance hit can be ~ 20% additional load

itslewiswatson:
@Dutchman101 the PR uses good old-fashioned procedural now

jlillis (approved):

ArranTuna (approved):

PR #367, opened 2021-11-03
deathpickups: some improvements

srslyyyy:
- Fixed pickup event handler (the code never passed through matchingDimension check because it was always nil).
- get methods are now called just on script init & when certain setting changes in resource.
- Improved readability & optimisation.

### DIFF
--- a/[gameplay]/deathpickups/deathpickups.lua
+++ b/[gameplay]/deathpickups/deathpickups.lua
@@ -4,9 +4,7 @@ local onlyCurrentWeapon = get("only_current")
 local dropRadius = get("radius")
 
 local function destroyDeathPickup(pickupElement)
-	local validPickup = isElement(pickupElement)
-
-	if validPickup then
+	if isElement(pickupElement) then
 		destroyElement(pickupElement)
 	end
 end
@@ -28,11 +26,7 @@ end
 
 function onDeathPickupHit(playerElement)
 	cancelEvent()
-
-	local weaponID = getPickupWeapon(source)
-	local weaponAmmo = getPickupAmmo(source)
-
-	giveWeapon(playerElement, weaponID, weaponAmmo, false)
+	giveWeapon(playerElement, getPickupWeapon(source), getPickupAmmo(source), false)
 	destroyDeathPickup(source)
 end
 addEventHandler("onPickupHit", resourceRoot, onDeathPickupHit)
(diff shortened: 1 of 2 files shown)


[gameplay]/deathpickups/deathpickups.lua
```diff
@@ -1,60 +1,109 @@
...
-		giveWeapon ( player, weapid, weapammo, false )
+local pickupTimers = {}
+local expireTime = get("timeout")
+local onlyCurrentWeapon = get("only_current")
+local dropRadius = get("radius")
+
+local function destroyDeathPickup(pickupElement)
+	local validPickup = isElement(pickupElement)
+
+	if validPickup then
```
Comment (jlillis) [inline [gameplay]/deathpickups/deathpickups.lua:9]: Can be shortened to `if isElement(validPickup) then`.

[gameplay]/deathpickups/deathpickups.lua
```diff
@@ -1,60 +1,109 @@
...
 end
 
-local function destroyDeathPickup ( pickup )
-	timers[pickup] = nil
-	removeEventHandler ( "onPickupHit", pickup, onDeathPickupHit )
-	destroyElement ( pickup )
+local function dropAllWeapons(posX, posY, posZ, droppedWeapons)
+	local weaponsCount = #droppedWeapons
+
+	for wID = 1, weaponsCount do
```
Comment (jlillis) [inline [gameplay]/deathpickups/deathpickups.lua:17]: Same as above, remove variable `weaponsCount` and just use `#droppedWeapons`

[gameplay]/deathpickups/deathpickups.lua
```diff
@@ -1,60 +1,109 @@
...
+	end
+end
+
+function onDeathPickupHit(playerElement)
+	cancelEvent()
+
+	local weaponID = getPickupWeapon(source)
+	local weaponAmmo = getPickupAmmo(source)
+
+	giveWeapon(playerElement, weaponID, weaponAmmo, false)
```
Comment (jlillis) [inline [gameplay]/deathpickups/deathpickups.lua:35]: Again, can be shortened by just using `getPickupWeapon(source)` and `getPickupAmmo(source)` rather than temporary variables.

jlillis (approved):
Looks good overall. Use of temporary variables makes this script longer than it needs to be, but perhaps more readable. Changes requested are minor and optional. Ready for merge later this week absent any other blocking comments.

[gameplay]/deathpickups/deathpickups.lua
Comment (srslyyyy) [reply [gameplay]/deathpickups/deathpickups.lua:17]: This is used below, so it's good to have it in variable instead of calling length check again.

Issue #365, opened 2021-11-02
Process initial Lua linter results & tweak its config


Dutchman101:
Lua linter from PR #354 (merged) will mostly be used to hint PR authors about potential mistakes & check off warnings before proceeding. But we also have some interesting results from the [initial scan](https://github.com/multitheftauto/mtasa-resources/runs/4078718561) of our entire repository.

Tasks:
* Identify the most common false positive warning types (due to MTA's Lua stack) so we can add more ignore rules, even after the last one we added: 7c135ae - the below warning log file will be helpful for this
* Triage the warning log to identify and fix issues with official resources in mtasa-resources
Warning log: [https://pastebin.mtasa.com/999998690](https://pastebin.mtasa.com/999998690) or see attached .txt file: [LuaLinter_results.txt](https://github.com/multitheftauto/mtasa-resources/files/7460010/LuaLinter_results.txt)

To work from: [Repository tree (version: 7c135ae)](https://github.com/multitheftauto/mtasa-resources/tree/7c135ae7970c4e6edc7adbe3c3a469340c0fd04d) to keep this log relevant for a while

Any help with either points will be appreciated. Especially skimming through warnings on our existing scripts and checking if they identified real problems, is time-consuming. If you have any candidates to be added to the ignore list, please comment them.

For a list of warning ID's and types see [luacheck/warnings.rst](https://github.com/mpeterv/luacheck/blob/master/docsrc/warnings.rst)

Disinterpreter:
In addertum to D101. All linters and static analysises not ideal.

So we recommend to listen them, but don't trust if the situation allows.

If you see the luacheck wrong in a specific situation, please use: https://luacheck.readthedocs.io/en/stable/inline.html?highlight=ignore#inline-options

Dutchman101:
> So we recommend listen to them, but don't trust if the situation allows

yes, that's why i said: skimming through warnings on our existing scripts and _checking if they identified real problems_, all static analyzer output requires manual checking as most of them are false positive/not relevant

iDannz1:
I've been checking for some time and so far i havent found any real problem... and IMHO we could add id `213` - `(unused loop variable)`, and maybe all `shadowing` warnings (`421`, `422`, `423`, `431`, `432`, `433`) to the ignore list.

iDannz1:
[...]

PR #368, opened 2021-11-04
acpanel: linter warning fix

ricksterhd123:
Fixed linter warnings for acpanel, not tested in-game yet.

### DIFF
--- a/[admin]/acpanel/s_img_mod.lua
+++ b/[admin]/acpanel/s_img_mod.lua
@@ -45,13 +45,13 @@ addEventHandler ( "onResourceStart", resourceRoot,
 function isImgModBlocked( name )
 	local defText = getBlockDefText()
 
-	local type = getPanelSetting( "blockmods.type" )
+	getPanelSetting( "blockmods.type" )
 	--outputDebug( "Checking " .. name .. " against type " .. tostring(type) )
 	--outputDebug( tostring(defText) )
 
 	local lineList = split(defText,"\n")
 	for _,line in ipairs(lineList) do
-		local line = string.gsub(line, "\r", "")
+		line = string.gsub(line, "\r", "")
 		--outputDebug( "Checking " .. name .. " against line " .. tostring(line) )
 		if line == "*" or string.find(name,line) then
 			return true
--- a/[admin]/acpanel/_common.lua
+++ b/[admin]/acpanel/_common.lua
@@ -35,7 +35,7 @@ function table.find(t, ...)
 		value = nil
 	end
 	for k,v in pairs(t) do
-		for i,index in ipairs(args) do
+		for _,index in ipairs(args) do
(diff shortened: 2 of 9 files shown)


Disinterpreter:
```
label1
label2
label3
```
Someone in the past didn't think about variables' name, should we set the another names?

iDannz1:
🤗 Well, thank you for this PR, but im not sure if we should start fixing lint warnings before we decide 100% about which rules we will use.... (see #365)

ricksterhd123:
I had that in mind, I thought it was be good used as reference until it was figured out. The issues in here were mainly unused variables or redeclaring previously declared variables.

Ricky
________________________________

ricksterhd123:
Proceed however you wish, it was originally a draft. If you feel it should be closed until #365 is resolved then go ahead. I'll be back ;)

Ricky
________________________________

iDannz1:
I mean, there are some chances we will end up removing or adding some new rule, and that might make you have to go back and re-edit this PR, but anyway, i believe the final answer should be yours, if you dont see any problem with that, i dont think you need to close, your time is yours 😁

Dutchman101:
@iDannz1 principally there is no problem with having PR's like this one, #365 is only meant to optimize the workflow against false positives _for ourselves_ (when it come to the linting when PR's with new features are submitted) and not to set any sort of style/coding guidelines

patrikjuvonen:
I think in general we want to apply automatic formatting and linting fixes, not focus on that stuff too much manually.

Dutchman101:
> I think in general we want to apply automatic formatting and linting fixes, not focus on that stuff too much manually.

yeah, that's also a good idea to look into as soon the linter config has been optimized for MTA's Lua stack (so that it only fishes out things with a high probability to indicate a real flaw or subpar coding conventions used), of course depending on how much we can adapt this linter for purposes like that and maybe also coding style enforcement once we got our guidelines in order as well. I will copy this feedback over.

ricksterhd123:
All the issues within this resource are mainly due to bad coding practices. I would argue that unused index in for loop could be ignored based on preference.

Ricky
________________________________

jlillis:
I think this sort of PR should wait until after discussion in #365 is concluded.

[admin]/acpanel/s_img_mod.lua
```diff
@@ -45,13 +45,13 @@ addEventHandler ( "onResourceStart", resourceRoot,
 function isImgModBlocked( name )
 	local defText = getBlockDefText()
 
-	local type = getPanelSetting( "blockmods.type" )
+	getPanelSetting( "blockmods.type" )
```
Comment (ArranTuna) [inline [admin]/acpanel/s_img_mod.lua:48]: If type was unused then surely this function call is redundant?

ArranTuna:
These changes have already been fixed or the warnings suppressed.

PR #372, opened 2021-11-17
Fix "previously defined" warnings in [admin]

ArranTuna:
Just fixed the "previously defined" warnings in [admin] and seeing how to make PR. I'm unsure as to the need for "previously defined" warnings to be fixed, some seem less necessary to be fixed than others.

### DIFF
--- a/[admin]/admin2/shared/utils.lua
+++ b/[admin]/admin2/shared/utils.lua
@@ -31,8 +31,8 @@ function RGBToHex(red, green, blue, alpha)
     end
 end
 
-function isAnonAdmin(player)
-    local player = (player or localPlayer)
+function isAnonAdmin(aplayer)
+    local player = (aplayer or localPlayer)
 
     if (not isElement(player)) then
         return false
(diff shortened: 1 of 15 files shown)


[admin]/admin2/shared/utils.lua
```diff
@@ -31,8 +31,8 @@ function RGBToHex(red, green, blue, alpha)
     end
 end
 
-function isAnonAdmin(player)
-    local player = (player or localPlayer)
+function isAnonAdmin(aplayer)
```
Comment (TheNormalnij) [inline [admin]/admin2/shared/utils.lua:34]: Also possible
```lua
function isAnonAdmin(player)
    player = (player or localPlayer)
```

TheNormalnij:
Looks good to me

[admin]/admin2/shared/utils.lua
Comment (ArranTuna) [reply [admin]/admin2/shared/utils.lua:34]: Your suggestion is exactly how it was before which causes this "previously defined" warning, that was 1 of the warnings where I don't think it helps to change.

Comment (TheNormalnij) [reply [admin]/admin2/shared/utils.lua:34]: This warning is triggered by additional local variable with same name as a previous defined variable. Correct me if I am wrong.
```lua
local a = 1
...
a = 3 -- Good
local a = 2 -- Warning
```
Function arguments are local variables as well

Comment (ArranTuna) [reply [admin]/admin2/shared/utils.lua:34]: Oh right you removed the local, my mistake.

ArranTuna:
I think I made this PR in the wrong way because I had to create a branch to create this PR and yet there aren't other branches for PR's so I'm guessing I have to clone the repository and make branches in the clone for PRs?

PR #377, opened 2022-02-06
admin: Remove redundant serial check

srslyyyy:
[Discord discussion](https://discord.com/channels/801330706252038164/801519095064428614/939557451638923314) (MTA dev server)
Confirmed by @Dutchman101 that it's no longer needed.

### DIFF
--- a/[admin]/admin/server/admin_server.lua
+++ b/[admin]/admin/server/admin_server.lua
@@ -285,6 +285,11 @@ function iif ( cond, arg1, arg2 )
 	return arg2
 end
 
+local serialExp = "^" .. string.rep ( "[A-F0-9]", 32 ) .. "$"
+function isValidSerial ( serial )
+	return serial:match ( serialExp )
+end
+
 function getWeatherNameFromID ( weather )
 	return iif ( aWeathers[weather], aWeathers[weather], "Unknown" )
 end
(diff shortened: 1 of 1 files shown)


Dutchman101:
I accidentally put my comment in the commit desc

But see it..

Serial validity (when it comes to characters) is guaranteed by the AC framework, so this check adds unnecesary overhead to a high traffic event, player connect.

In the unlikely event that AC goes out of support in decades, this PR has to be reverted and server owners notified.

I should remark that the check didn't even work properly, as proven during an incident 2 years ago when a cheat developer managed to get a totally invalid serial made up of weird characters.. which was read as such by the server using getPlayerSerial, but didn't trigger a kick. So if this PR ever gets reverted, `local serialExp = "^" .. string.rep ( "[A-F0-9]", 32 ) .. "$"` has to be investigated

Dezash:
A bit late to the party, but this has also removed serial validation when adding a serial ban in the bans tab.
Entering invalid value will not output any error as it did before with
`outputChatBox ( "Error - Invalid serial", source, 255, 0, 0 )`
image

@Dutchman101 the regex was working perfectly fine, it was fixed on Aug 7, 2019 here: https://github.com/multitheftauto/mtasa-resources/pull/164

PR #378, opened 2022-02-06
admin2: Remove redundant serial check.

srslyyyy:
[Same as in here.](https://github.com/multitheftauto/mtasa-resources/pull/377)

### DIFF
--- a/[admin]/admin2/server/admin_server.lua
+++ b/[admin]/admin2/server/admin_server.lua
@@ -325,12 +325,7 @@ addEventHandler(
                 end
             elseif (action == "banserial") then
                 mdata = data
-                if (isValidSerial(data)) then
-                    if (not BanSerial(string.upper(data), source)) then
-                        action = nil
-                    end
-                else
-                    outputChatBox("Error - Invalid serial", source, 255, 0, 0)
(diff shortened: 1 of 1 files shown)


ArranTuna:
It stops you entering an invalid length serial how is this redundant?

srslyyyy:
@ArranTuna how it can stop it if the function doesn't exist there?
obraz

Anyways it's my overlook, because this check was present in admin during player join, see mentioned PR.

ArranTuna:
Ok so yeah it's not in admin2 but it's in admin so really shouldn't you have made a PR to add

```
local serialExp = "^" .. string.rep ( "[A-F0-9]", 32 ) .. "$"
function isValidSerial ( serial )
	return serial:match ( serialExp )
end
```

to admin2 instead of removing it from admin and admin2?

srslyyyy:
The whole discussion was about removing it from admin, because it was confirmed that serial check (during player join) was redundant, because MTA AC will take care of it anyway (serial manipulating). I guess it can be added for both admins in separate PRs (feel free to do so). So possibly i might close those two PRs or leave the function at least in (#377) (for serial ban check).

PR #381, opened 2022-03-09
ctf: maintenance (fixes, cleanup & refactor)

IIYAMA12:
- Cleanup + refactor.
- Errors/warnings remove.
- For team selection including also the keys arrow up/down. For a more user friendly experience.
- Fixing a broken map script.

### DIFF
--- a/[gamemodes]/[ctf]/ctf/ctf.lua
+++ b/[gamemodes]/[ctf]/ctf/ctf.lua
@@ -392,7 +392,7 @@ function CTF_flag_check ( )
 end
 
 function CTF_onColShapeHit ( player )
-	if ( ( getPlayerName( player ) ~= false ) and ( isPedDead ( player ) == false ) ) then
+	if ( (getElementType(player) == "player") and ( getPlayerName( player ) ~= false ) and ( isPedDead ( player ) == false ) ) then
 		local playerTeam = getPlayerTeam( player )
 		local playerCol = getElementData( player, "col" )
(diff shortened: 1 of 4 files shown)


IIYAMA12:
Ready for review.

[gamemodes]/[ctf]/ctf/ctf.lua
```diff
@@ -134,6 +131,7 @@ function CTF_gamemodeMapStart ( startedMap )
 				for flagKey,flagValue in ipairs(flags) do
 					local x,y,z = tonumber( getElementData ( flagValue, "posX" ) ), tonumber( getElementData ( flagValue, "posY" ) ), tonumber( getElementData ( flagValue, "posZ" ) )
 					local object = createObject( 2993, x, y, z )
+					setElementCollisionsEnabled ( object, false )
```
Comment (Proxy-99) [inline [gamemodes]/[ctf]/ctf/ctf.lua:134]: there is no need to call the function setElementCollisionsEnabled
the object has no collision even though it is true there is no collision

Comment (IIYAMA12) [reply [gamemodes]/[ctf]/ctf/ctf.lua:134]: It is disabled because the flag can be broken. This happens when it is detached from the player (probably breaks because it gets stuck inside of the player) and it breaks because of explosions.
When collision is disabled, it can't break.

IIYAMA12:
Hey @jlillis, do you perhaps have some time to review my pull request?

Dutchman101 (approved):
Looks good, thanks

Issue #358, opened 2021-10-24
Resourcemanager is broken


Xenius97:
**Which resource(s) have this problem?**
resourcemanager

**Describe the bug**
Resourcemanager stopped working after browser update

**To Reproduce**
Download latest resources

**Screenshots**
https://i.imgur.com/w9QAY0z.png

**Version**
Latest resource pack + Chrome 95.0.4638.54 / Tested also on latest firefox, not working

ArranTuna:
I have latest resources and not getting any errors when restarting it:

![image](https://user-images.githubusercontent.com/253404/142212690-05b19d8c-09e8-4551-84f1-0bbfbfb30ac3.png)

iDannz1:
I was going to create a new issue about this bug right now but just found this one....

@ArranTuna steps to reproduce:
1. start the resource `resourcemanager`.
2. go to http://localhost:[http-port]/resourcebrowser/.
3. select any resource in the list and see the server console.

I took a little look at the code and noticed that this is happening because the [getResourceInfo](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Bweb%5D/resourcemanager/manager.lua#L21) function expects a resource as parameter, but it's receiving a table with the keys `name` and `state`.

Xenius97:
Something is broken in javascript, sorry I didn't see your answer Arran, didn't get email.
Select a resource, and you can't even do anything because of these error.

ArranTuna:
I've fixed it. I guess at some point some code change some where (unless it never worked to begin with) caused a behaviour change that broke this. If you pass a resource userdata like getResourceFromName("acpanel") javascript receives {"name":"acpanel"} instead of a userdata but this script acted as if the resourcedata was passed correctly but then it also seems to use this .name thing so I dunno it's weird, but everything works now so yeah...

PR #383, opened 2022-04-07
Remove legacy json lib

### LINKED ISSUE #379: elementbrowser is broken

elementbrowser

**To Reproduce**

1 - start the `elementbrowser` resource
2 - open http://server_ip:http_port
3 - select the "Element browser" menu option

**Expected behavior**

it should work 😄

**Screenshots**

image
image

**Version**

MTA:SA Server 1.5.9-release-21048

**Additional context**

I haven't checked the code, but I belive its the same thing @ArranTuna fixed in [this commit](https://github.com/multitheftauto/mtasa-resources/commit/37348cecb60a6c999c316398a17f866d54569953)

4O4:
Last modifications made to `json.js` lib during webmap upgrade ([[web]/ajax/json.js](https://github.com/multitheftauto/mtasa-resources/commit/8e2b12b6d012c89a063a8c5e9873454f8d3ab7e3#diff-adfe7968957f2a79a6ec34240a7f5d87186a3b1a64b8c52baa91f1f78c5c7a29)) broke other resources. This lib however is no longer needed as there have been native JSON functions in Javascript for ages already.

This PR reverts a workaround previously introduced in #358, removes the old lib and replaces it with native JSON functions. 

This also involved a slight refactoring. A bit of more modern code was introduced (`class`es, `startsWith`). I'm assuming we don't care about ancient browsers anymore and ES2015 is a safe target for any new JS code.

Closes #379 #358


[web]/ajax/ajax.js
```diff
@@ -115,18 +115,16 @@ function getKeyCode ( e )
...
-		for ( i = 0; i < this.elements.length; i++ )
-		{
+class ElementManager {
+	constructor() {
+		this.elements = new Array();
+	}
+	
+	get(id) {
+		for (let i = 0; i < this.elements.length; i++) {
 			var element = this.elements[i];
```
Comment (patrikjuvonen) [inline [web]/ajax/ajax.js:125]: Prefer [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).

```suggestion
		for (const element of this.elements) {

```

[web]/ajax/ajax.js
```diff
@@ -137,18 +135,16 @@ function ElementManager() {
...
+	}
 
-function ResourceManager() {
-	this.resources = new Array();
-	this.get = function (name) {
-		for ( i = 0; i < this.resources.length; i++ )
-		{
+	get(name) {
+		for (let i = 0; i < this.resources.length; i++) {
 			var resource = this.resources[i];
```
Comment (patrikjuvonen) [inline [web]/ajax/ajax.js:145]: Prefer [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).
```suggestion
		for (const resource of this.resources) {
```

[web]/elementbrowser/script.js
```diff
@@ -26,10 +26,7 @@ function loadElementInfo(element)
 			var dataList = "{";
 			for ( var key in data )
 			{
-				if ( key != "toJSONString" )
-				{
-					dataList += key + " = " + data[key] + ", ";
-				}
+				dataList += key + " = " + (typeof(data[key]) === "object" ? JSON.stringify(data[key]) : data[key]) + ", ";
```
Comment (patrikjuvonen) [inline [web]/elementbrowser/script.js:29]: This for loop should be refactored to use [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).

[web]/resourcemanager/manager.lua
```diff
@@ -37,31 +30,3 @@ function getResourceInfo ( resource )
...
-	end
-	mta_stopResource(resource)
-end
-
-function restartResource(resource)
-	if (type(resource) == "table") then
-		resource = getResourceFromName(resource.name)
-	end
-	mta_restartResource(resource)
-end
```
Comment (patrikjuvonen) [inline [web]/resourcemanager/manager.lua:67]: By removing these functions, are we breaking backwards compatibility for the exported http functions?
https://github.com/4O4/mtasa-resources/blob/bugfix/remove-legacy-ajax-json-lib/%5Bweb%5D/resourcemanager/meta.xml#L8-L16

patrikjuvonen (changes requested):
Requesting for clarification at least on the http exports part.

[web]/resourcemanager/manager.lua
Comment (4O4) [reply [web]/resourcemanager/manager.lua:67]: Nope, nothing gets broken. These were just thin wrappers added to workaround the problem. Without them the exported functions are just native MTA ones (so native `startResource` and so on).

patrikjuvonen (approved):

patrikjuvonen:
Thank you.

PR #389, opened 2022-05-31
editor_main: Disable OOP in meta.xml for scripting extensions

srslyyyy:
Disabling it due of being unnecessary (no single OOP call was being used).


jlillis (commented):
I'm inclined to approve but wondering if this could introduce a minor reversion in behavior. By disabling OOP for scripting extensions now when it was enabled previously, extensions that were since written using OOP will stop working by default. Thoughts?

PR #391, opened 2022-08-15
Fix "No map selected!" message box when clicking search map editbox

Mkl21:
extra check added to client/gui/admin_maps.lua line 69 to make exception if the source is the search bar editbox.


Mkl21:
I realized there is more cases to handle (everytime you cliked an element when nothing is selected in the gridlist) so this PR is not pertinent for the moment.

PR #393, opened 2022-12-04
New Feature: Car Steering #Fixed

francisco-create:
**In the update I made, it will remain as you left the wheel while getting out of the vehicle.**


Dutchman101:
I'm sorry, your contribution (and the way you submitted it) doesn't meet the standards of this project. I also don't think this resource has been authored by you.

Thanks for the interest though!

PR #395, opened 2022-12-08
freecam: Add support for changing field of view

Xenius97:
This allows you to change fov from MTA settings. Helps a lot of when playing on wide-screen or editing map.


lopezloo:
Maybe it would be better to not call `dxGetStatus()` every frame?

Xenius97:
@lopezloo check latest commit 51313370ef4aa5c176713a1325c52db64594d59c :)

PR #397, opened 2022-12-08
ajax: add json detection

### LINKED ISSUE #392: Use application/json in ajax resource
ajax

**Describe the bug**
The ajax resource is using Content-Type `application/x-www-form-urlencoded` instead of `application/json`  when sending JSON data in the request body. The embedded web server parses the JSON data as if it were a bunch of key=value pairs.

https://github.com/multitheftauto/mtasa-resources/blob/3e4e3a4b8c3bf554b2d2b71ea987dc6ffec3be2b/%5Bweb%5D/ajax/ajax.js#L90

**To Reproduce**
Inspect the http traffic when browsing the MTA server's web server.

**Expected behavior**
Ajax uses `application/json`.

Xenius97:
resolves #392

### DIFF
--- a/[web]/ajax/ajax.js
+++ b/[web]/ajax/ajax.js
@@ -26,6 +26,15 @@ function decode( uri ) {
     return uri;
 }
 
+// Check data is JSON
+function isJSON(str) {
+	try {
+		JSON.parse(str);
+	} catch (e) {
+		return false;
+	}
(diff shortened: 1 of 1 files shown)


patrikjuvonen (approved):

patrikjuvonen:
Thank you.

PR #400, opened 2023-01-04
editor: Rotation improvements

Zangomangu:
- Use quaternions instead of euler angles for incremental rotations, used with mouse scroll and keyboard. 
- Add ability to rotate elements in local space (Default: Rctrl)

https://user-images.githubusercontent.com/59518113/210654480-59087802-7160-4cb6-be39-56ba88b1bb73.mp4

- Adds an xml with offsets for 150 default objects with skewed rotations to make them axis aligned (Thanks to Tut). The offsets are applied first time rotating the object, making them easier to rotate.

- The offset feature is toggled with rotation patch option (default: enabled). Also changed text of below options to better describe what they do.
image


Zangomangu:
done

Issue #374, opened 2021-12-24
editor: add some way to restore removed objects


iDannz1:
**Is your feature request related to a problem? Please describe.**
Apparently at this moment there is no way to restore objects that were removed from the original map... (we have to manually find and remove the correct line in the .map file)

**Describe the solution you'd like**
As we can list the changes through this window, maybe we could add some button to restore the removed object?
![image](https://user-images.githubusercontent.com/38073795/147357644-afe0db67-795f-4b6c-ac85-198c25a7e90e.png)
![image](https://user-images.githubusercontent.com/38073795/147357506-07f988a1-46bc-48c8-8e6b-2798d2d1f61e.png)
(we can also add navigation using the keyboard arrows to the above list, as is already possible in other lists, such as the list of objects/ids)

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Vampiiree:
Actually, there's a way of restoring world objects using the 'current elements' tool:
- Click twice in the line so it opens the properties
- There will be a symbol (>) on the lower right side
- Press 'delete'

https://i.imgur.com/ZNsnE24.png

psyhiic:
> Actually, there's a way of restoring world objects using the 'current elements' tool:
> 
> * Click twice in the line so it opens the properties
> * There will be a symbol (>) on the lower right side
> * Press 'delete'
> 
> https://i.imgur.com/ZNsnE24.png

But perhaps he gave a more practical idea of ​​restoring the removed object.

Vampiiree:
Well, I didn't say that his idea isn't good. I just pointed out that it already exists within the tool he mentioned.

Fernando-A-Rocha:
+1

Button to restore removed world object from the list + delete key to restore + navigating with arrow keys

All good enhancements

iDannz1:
> Actually, there's a way of restoring world objects using the 'current elements' tool:
> 
> * Click twice in the line so it opens the properties
> * There will be a symbol (>) on the lower right side
> * Press 'delete'
> 
> https://i.imgur.com/ZNsnE24.png

Yeah, I didn't know about that...

psyhiic:
> > Actually, there's a way of restoring world objects using the 'current elements' tool:
> > 
> > * Click twice in the line so it opens the properties
> > * There will be a symbol (>) on the lower right side
> > * Press 'delete'
> > 
> > https://i.imgur.com/ZNsnE24.png
> 
> Yeah, I didn't know about that...

[...]

Issue #301, opened 2021-03-12
editor: Inconsistency in destroying elements


Yamsha75:
**Which resource(s) have this problem?**
editor_main

**Describe the bug**
When destroying an element using editor_gui (for example with DELETE button), first `"onElementDestroy"` event is triggered, and then the actual action is made, which is in fact just moving the element from working dimension (200) to destroyed elements dimension (201), as seen here:
https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/editor_main/server/createdestroy.lua#L99-L102
https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/editor_main/server/undoredo_action.lua#L210

After this action, when doing an UNDO and then a REDO, the order is different: first the element's dimension is changed and then the event is triggered, as seen here:
https://github.com/multitheftauto/mtasa-resources/blob/master/%5Beditor%5D/editor_main/server/undoredo_action.lua#L258-L260

**To Reproduce**
1. Run this code in runcode:
```lua
srun function f(e) outputChatBox(e.dimension) end addEventHandler("onElementDestroy", root, function() f(source) setTimer(f, 50, 1, source) end)
```
2. Create an element in editor
3. Delete this element - observe "200" and "201" in chat - this shows that at onElementDestroy the event was still in working dimension and then moved to destroyed elements dimension
4. UNDO
5. REDO - observe "201" twice in chat - this time at onElementDestroy event the element was already in destroyed elements dimension

**Expected behavior**
This behaviour should be more consistent, as the only way to check if an editor element is destroyed or not is checking if its dimension is 200 or 201.

**Screenshots**

**Version**
Multi Theft Auto v1.5.8-release-20802

**Additional context**

Haxardous:
fixed by #401

PR #401, opened 2023-01-13
Add a new "restore" button within Editor's "Current Elements" window.

Haxardous:
Although this commit is for #374 -- it also fixes the problems in #301 (without major changes to object deletions (for not e.g. replacing the dimension method with something else))

Knowing that this commit is also compatible to redo/undo actions*
image


Fernando-A-Rocha:
approved ✔️

PR #385, opened 2022-04-18
editor_main: Improve scripting extensions

srslyyyy:
Makes use of onPlayerResourceStart in order to send data to client.

PR #408, opened 2023-01-22
editor_main: Fix version warning

srslyyyy:
This PR fixes version warning introduced in #385 

obraz

reported by @brenodanyel


jlillis (changes requested):
I think we ought to fetch the min_mta_version used by the editor resource so we're not hunting for this line of code each time we need to bump the version.

jlillis (approved):

PR #414, opened 2023-02-17
ip2c: Add safety checks to exported function arguments

Fernando-A-Rocha:
Adds safety checks to arguments of `getCountryName( country )` and `getPlayerCountry ( player )`.

Also fixes `admin2` calling `getCountryName(aPlayers[player].country)` without checking if the country code is set. This would always trigger an error when playing on LAN as ip2c can't get any country from your local ip.

### DIFF
--- a/[admin]/admin2/server/admin_server.lua
+++ b/[admin]/admin2/server/admin_server.lua
@@ -154,7 +154,9 @@ function updatePlayerCountry(player)
     local isIP2CResourceRunning = getResourceFromName( "ip2c" )
 	isIP2CResourceRunning = isIP2CResourceRunning and getResourceState( isIP2CResourceRunning ) == "running"
     aPlayers[player].country = isIP2CResourceRunning and exports.ip2c:getPlayerCountry(player) or false
-    aPlayers[player].countryname = isIP2CResourceRunning and exports.ip2c:getCountryName(aPlayers[player].country) or false
+    if aPlayers[player].country then
+        aPlayers[player].countryname = isIP2CResourceRunning and exports.ip2c:getCountryName(aPlayers[player].country) or false
+    end
 end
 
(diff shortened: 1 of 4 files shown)


Fernando-A-Rocha:
I noticed that on a new server with default resources, ip2c doesn't have fetchRemote access so I added the aclrequest to meta.xml

e

Was this the right thing to do?

brenodanyel:
> I noticed that on a new server with default resources, ip2c doesn't have fetchRemote access so I added the aclrequest to meta.xml
> 
> e
> 
> Was this the right thing to do? Is there a way to allow it by default?

Maybe it's better to send a PR on [mtasa-blue](https://github.com/multitheftauto/mtasa-blue/blob/4aabd6e75b454106a1d1db2c84fbaa3c10537bf0/Server/mods/deathmatch/acl.xml#L18) adding a simple ``<object name="resource.ip2c"/>`` under the "Admin" group to enable it by default, so that server owners don't have to worry about running `aclrequest allow ...` manually?

Fernando-A-Rocha:
> > I noticed that on a new server with default resources, ip2c doesn't have fetchRemote access so I added the aclrequest to meta.xml
> > e
> > Was this the right thing to do? Is there a way to allow it by default?
> 
> Maybe it's better to send a PR on [mtasa-blue](https://github.com/multitheftauto/mtasa-blue/blob/4aabd6e75b454106a1d1db2c84fbaa3c10537bf0/Server/mods/deathmatch/acl.xml#L18) adding a simple `<object name="resource.ip2c"/>` under the "Admin" group to enable it by default, so that server owners don't have to worry about running `aclrequest allow ...` manually?

I was looking for this. Thank you, this seems the right thing to do. I'll revert my latest aclrequest commit and do that instead 👍

PR #416, opened 2023-02-27
admin2: fixed various sorting related gridlist issues

Dark-Dragon:
_guiGridListSetItemText_ and _guiGridListSetItemData_ can affect unintended rows when user-sorting is used in conjunction with the search-filter, causing the text or data to apply to an incorrect row and therefore disappear for the intended row with the current way admin2 refreshes the gridlists. In some cases it's a big deal, because it can cause the gridlist to malfunction until admin2 is reloaded. Rather than disabling user-sorting for all gridlists we can use the optional arguments of _guiGridListAddRow_ wherever possible, and store, disable and re-apply the GUI property "SortDirection" for the remaining cases, to ensure proper application.

This pull request currently fixes the following GUI gridlists:

_/client/main/admin_players.lua_ -> **aPlayersTab.PlayerList**
_/client/main/admin_resources.lua_ -> **aResourcesTab.ResourceList**
_/client/widgets/admin_interior.lua_ -> **aInterior.List**
_/client/widgets/admin_skin.lua_ -> **aSkin.List**
_/client/widgets/admin_team.lua_ -> **aTeam.List**
_/client/widgets/admin_vehicle.lua_ -> **aVehicle.List**
_/client/widgets/admin_weapon.lua_ -> **aWeapon.List**

Notably it's currently not changing these GUI gridlists:

_/client/main/admin_acl.lua_ -> **aAclTab.RightsTab** (Benefit from changes would be diminished, as clearing all columns (which is done a lot here) circumvents the problem as well. Main remaining benefit would be to preserve user-sorting after input in the search-filter)
_/client/main/admin_bans.lua_ -> **aBansTab.BansList** (Bans are currently a large WIP in another branch, so I decided to not change this one for now)

Three small additional fixes were made: one removing tabs in favor of 4 spaces to match coding style in admin_players.lua; fixed an incorrectly named setting to have admin2 remember the users sort vehicles by vehicle groups preference; and the team widget will now close correctly when assigning a player a team via double click.

See line comments on commits for more explanation.


jlillis:
Looks good from my testing, nice work.

PR #420, opened 2023-04-03
hay: fixed #346 Hay doesn't reset

### LINKED ISSUE #346: Hay doesn't reset
When a game of hay is over (player picks up the burger) it should teleport everyone down and re-create the burger pickup but it doesn't.

It should teleport everyone down and re-create the burger pickup.

-

**Additional context**
I don't have any coding experience so I wasn't able to inspect the code but I've checked debugscript monitor while picking up the burger and nothing came up. However, there was an outputChatbox that was getting triggered when I picked up the burger, so script knows I won but it doesn't reset it. (or even if it does, it's not resetting fast enough.) I tried jumping down and respawning but no luck.

I have a test server set up, feel free to test at 195.174.168.123 (superman is added and should be available to everyone as well. just type /superman )

Dark-Dragon:
The hay gamemode will now start a new round 15 seconds after a player reaches the top, as requested in #346 .

Feedback appreciated, specifically about the following:

- Is 15 seconds good?
- Pickup is no longer a minigun, so do we even need to toggle the "fire" control for the winner?
- Do we want to put effort into having a set number of rounds play out before a gamemode vote starts?


jlillis:
>     * Is 15 seconds good?
I'd say 10 is a good default, but might as well make it a resource setting.

>     * Pickup is no longer a minigun, so do we even need to toggle the "fire" control for the winner?
Maybe the pickup *should* be a minigun?

>     * Do we want to put effort into having a set number of rounds play out before a gamemode vote starts?
When I refactored `deathmatch` I used `mapcycler` to handle changing the map/mode at the end of the round (if it was running). Not sure if you can do that here since the `hay` resource does not use maps - the version I was working on was going to.

Dark-Dragon:
We can re-add the minigun, I'm not opposed to the idea. If it was originally removed because it would leave the gamemode in an unplayable state for everyone else, then this would be a good moment to add it back.

However if we add the minigun back, I'd say 10 and even 15 seconds to wreak havoc with it seems a little short. We can definitely turn it into a setting, but I'd argue we would need to default to at least ~30 seconds. Still not nearly enough to get to the ground level, but at least you should be able to kill some of your competitors, as you make your way back down. Otherwise I just don't think it's going to feel very rewarding.

Thoughts?

Dark-Dragon:
> Could this possibly be from players falling through the map, as if they hadn't loaded that area of the map yet?

It's hard to imagine. The resource creates a colCircle and uses the onColShapeLeave event to kill players. From my understanding a colCircle should extend infinitely up and down, so there should never be a case where players are too far up or down, and I don't see how the player could ever be considered outside of that col shape if they start within, end within and any position in-between the start and end point is also within the col shape.

One noteworthy thing is that when the resource starts it creates this col circle with a delay of 1 second, so that makes me think that this exact thing might have been a problem before, at least on resource start. From my testing respawning after dying to the col circle never seemed to result in another immediate death after spawn.

The only other thing I could see causing this would be player position extrapolation. If for some reason the server considered a position change from the player at the moment of spawn as movement that needs to be extrapolated to compensate for any delays, then it would make sense that the col circle wants to kill them, if the extrapolated position extends to outside of the col circle.

jlillis:
I think for now we'll consider the spawn issue something that will get taken care of in @MegadreamsBE's rewrite and get this merged. I can't reproduce it very often and don't think it's worth going further down the rabbit hole if it's all getting rewritten anyway.

PR #429, opened 2023-04-24
admin2: fixed /report message viewer widget

Dark-Dragon:
Currently the widget to view messages sent via the /report command form is barely functioning. It will only display messages in the gridlist, with no way to to view them or delete them. Additionally some oversight in admin_storage.lua results in messages creating duplicate copies of themselves whenever you restart admin2.

This pull request addresses these issues and removes the now redundant admin_message.lua


jlillis (approved):
Great work. I took this opportunity to adjust some UI size issues I noticed in this widget in an added commit.

PR #433, opened 2023-06-18
Update label for new fps limit

vyn666:


jlillis:
Thanks for this PR. I found other things that needed to be updated in `admin2` for the new FPS limit changes and did them via additional commits here.

PR #439, opened 2023-07-07
helpmanager: removed redundant invisible window

Dark-Dragon:
It's empty, invisible at all times and leaves you confused if you accidentally click it which brings it to front and therefore makes any other gui elements underneath seem unresponsive. I can't think of what purpose this thing might have (had).


jlillis:
I think it was intended to ensure the helpmanager was "AlwaysOnTop" - I've used the same trick before I know about the CEGUI property that does the same thing. Not sure why it needs to be always on top. If nobody else can think of a reason I think this is good to merge.

jlillis (approved):

Dark-Dragon:
I see. In it's current state however, if you were to click it, it would prevent any sort of interaction with the helpmanager itself as well up until you close and re-open it. Additionally, if you know this window exists (or by pure chance) you can move the window out of the way, so it's not fool-proof either. I'm definitely in favor of removing it, but if we want it to stay, we need to fix the annoyances it currently causes.

PR #447, opened 2023-08-15
Update skins.xml

F2BShady:
In version 1.6, I realised that the new special skins added in the freeroam skins were not on the freeroam skins side and I made additions, I would like you to edit some of the skins because I don't know the names of some skins, as a test I saved some names as "NAME"


F2BShady:
yes, I made an arrangement

PR #457, opened 2023-11-23
Add runcode aclrequest for loadingstring

IIYAMA12:
Currently runcode does not work in 1.6 because it uses loadstring, which is disabled.
For the developer is handy to give loadstring rights with aclrequest, making sure it is less likely that the user will grand the resource admin permissions.


Allerek:
Are you sure loadstring is disabled? Im sure @thisdp uses it in his DGS project and it works fine https://github.com/thisdp/dgs

IIYAMA12:
> Are you sure loadstring is disabled? Im sure @thisdp uses it in his DGS project and it works fine https://github.com/thisdp/dgs

I am pretty sure.

See acl.xml in mta-blue:
https://github.com/multitheftauto/mtasa-blue/blob/abec07aeb90eef19ed47396baed3b190994be54a/Server/mods/deathmatch/acl.xml#L141

And commit:
https://github.com/multitheftauto/mtasa-blue/commit/89e2d375d12deb026ee91fedc5e1ced04dc9a723

Anyway, very good that this function is disabled by default.

thisdp:
> Are you sure loadstring is disabled? Im sure @thisdp uses it in his DGS project and it works fine https://github.com/thisdp/dgs

ah, clientsided loadstring can affect DGS. But this pr is for serversided loadstring.

TheNormalnij:
Thanks you

PR #458, opened 2023-11-24
Fix #455

### LINKED ISSUE #455: Remove or load admin_definitions.lua
admin

**Describe the bug**
[admin_definitions.lua](https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/admin/admin_definitions.lua) is not in the meta.xml and is therefore never loaded. Now currently the admin resource seems to only have minor issues without it, but having the file around but not loading it causes confusion. For instance #440 didn't properly fix #373, because it assumed the file was loaded.

It should therefore be decided whether the file can be safely removed (I think it's only essential in the admin2 resource) or implemented properly. At the very least the ADMIN_CHAT_MAXLENGTH global variable needs to be set elsewhere if the file is deemed redundant.

**To Reproduce**
1. start admin
2. see debug warning relating to ADMIN_CHAT_MAXLENGTH being nil
3. find out it makes sense because the file which defines it is never loaded

**Version**
Client/Server 1.6
admin 1.5.9 (latest at the time of post despite version discrepancy)

jlillis:
Fixes #455 by loading admin_definitions.lua in `admin`. Also cleans up the file to remove unused definitions.


Dark-Dragon:
This pull request should probably also remove the following line: https://github.com/multitheftauto/mtasa-resources/blob/666246579c2ec596303614c3bebb6722e6674d67/%5Badmin%5D/admin/client/admin_client.lua#L13

[admin]/admin/meta.xml
```diff
@@ -8,6 +8,7 @@
              versions of the admin system.
   -->
   <!--Main script files, keep them in that order or it may break-->
+  <script src="admin_definitions.lua" type="shared" cache="false" />
```
Comment (TheNormalnij) [inline [admin]/admin/meta.xml:11]: No reasons for "cache="false""

Comment (Dutchman101) [reply [admin]/admin/meta.xml:11]: > No reasons for "cache="false""

The idea is to leave no clientside .lua files (see 6662465) for admin, who knows what this file might contain in the future

PR #461, opened 2024-01-16
Improved chat + Fix #460 (Chat messages sent twice)

### LINKED ISSUE #460: Chat messages sent twice
Playercolors

**Describe the bug**

**To Reproduce**
2. Send a message in the chat

**Expected behavior**

**Screenshots**
image

**Version**
Client OS: Windows 11

Server OS: Windows 11

**Additional context**
-

Nico8340:
A completely new and configurable solution has been added to admin and admin2.

It is possible to turn off Anti-Spam, edit monitoring and suspension times, disable word repetition, enable linking with playercolors and allow players to use custom HEX color codes. (And it fixes #460)


0xHexadecimal:
This breaks playercolors if you want to use it independently of freeroam. The reason chat messages are sent twice now is because playercolors got added to mtaserver.config as a default starting resource where this wasn't this case before.

Nico8340:
> This breaks playercolors if you want to use it independently of freeroam. The reason chat messages are sent twice now is because playercolors got added to mtaserver.config as a default starting resource where this wasn't this case before.

I misunderstood things a bit, now I went over everything again. It is true that if we want to use it without freeroam, it will not work properly because of this commit. However, the solution is not to remove it from mtaserver.conf, since playerblips needs the playercolors resource to work properly.

Nico8340:
Now playercolors work properly even if freeroam is not running.

0xHexadecimal:
> I misunderstood things a bit, now I went over everything again. It is true that if we want to use it without freeroam, it will not work properly because of this commit. However, the solution is not to remove it from mtaserver.conf, since playerblips needs the playercolors resource to work properly.

having playercolors running does nothing other than suppress a warning from playerblips. It might make more sense to change playerblips to ignore the running of playercolors if freeroam is running or remove the functionality from freeroam that already assigns random player nametag colours.

Nico8340:
> having playercolors running does nothing other than suppress a warning from playerblips. It might make more sense to change playerblips to ignore the running of playercolors if freeroam is running or remove the functionality from freeroam that already assigns random player nametag colours.

In this case, I think it is simpler and more meaningful to modify playerblips, since freeroam's onPlayerChat event handler contains other things, such as word repetition checking and spam protection. For this reason, playercolors must be removed again from mtaserver.conf.

Nico8340:
The changes have been made, if they are accepted in their current form, a pull request will be created in the main repository regarding the removal of the playercolors resource from mtaserver.conf.

jlillis:
The proper fix (imo, which I was working on) is the following:

- Remove all `onPlayerChat` handlers from `playercolors` and `freeroam`
- Re-implement their logic in `admin` and `admin2`, providing settings to control behavior (anti-spam parameters, color codes, etc)
- (optional) Submit PR to `mtasa-blue` to make the chatbox respect nametag colors by default.

Nico8340:
> The proper fix (imo, which I was working on) is the following:
> 
> * Remove all `onPlayerChat` handlers from `playercolors` and `freeroam`
> * Re-implement their logic in `admin` and `admin2`, providing settings to control behavior (anti-spam parameters, color codes, etc)
> * (optional) Submit PR to `mtasa-blue` to make the chatbox respect nametag colors by default.

It's ok, I'll work on the solution.

Nico8340:
The new implementation is ready, the onPlayerChat event handlers have been removed from all gamemodes -- A completely new and configurable solution has been added to admin and admin2.

It is possible to turn off Anti-Spam, edit monitoring and suspension times, disable word repetition, enable linking with playercolors and allow players to use custom HEX color codes.

Dutchman101:
It would be nice to use this opportunity for blocking common spam charsets (like Bismillah)

Dutchman101:
It would be nice to use this opportunity for blocking common spam characters (like Bismillah).. there is a way to select by charset to stop weird characters like that, if you don't know how i will post it soon

Nico8340:
> It would be nice to use this opportunity for blocking common spam characters (like Bismillah).. there is a way to select by charset to stop weird characters like that, if you don't know how i will post it soon

How do you imagine the implementation? Ban certain keywords or limit the message to a certain character set? (a-Z, 0-9, etc)

0xHexadecimal:
Is this not an even more inconvenient solution, as now if people aren't running freeroam and using the latest version of admin it will mess with chat and may cause issues like chat duplication for people using their own chat system? Now people will have the problem wonder why its happening then will have to edit admin to have what was normal behavior before.

Nico8340:
> Is this not an even more inconvenient solution, as now if people aren't running freeroam and using the latest version of admin it will mess with chat and may cause issues like chat duplication for people using their own chat system? Now people will have the problem wonder why its happening then will have to edit admin to have what was normal behavior before.

I will add a setting option to turn off the chat. The admin and admin2 resources are primarily designed for the default mta resources.

0xHexadecimal:
> I will add a setting option to turn off the chat. The admin and admin2 resources are primarily designed for the default mta resources.

Even then I don't think admin should have its own chat system. It's not something I'd expect admin to do. Having it block/remove certain keywords or prevent spamming is one thing but using it to have coloured names in chat is another.

jlillis:
> Even then I don't think admin should have its own chat system.

Strongly disagree. Anti-spam is a basic security task and the admin panel is the best place to do it. It also happens to be a convenient place to add the logic needed to handle player nametag colors and color codes in chat rather than having it re-implemented multiple times across other resources. If server owners already have their own systems in place, they can turn these features off.

[admin]/admin/conf/chat.xml
```diff
@@ -0,0 +1,9 @@
+<chat>
```
Comment (jlillis) [inline [admin]/admin/conf/chat.xml:1]: The configuration files in admin[2]/conf are mainly for static data rather than actual configuration settings. Consider using `get` or `set` instead.

Nico8340:
> Strongly disagree. Anti-spam is a basic security task and the admin panel is the best place to do it. It also happens to be a convenient place to add the logic needed to handle player nametag colors and color codes in chat rather than having it re-implemented multiple times across other resources. If server owners already have their own systems in place, they can turn these features off.

I agree with your opinion. Do you have any ideas on how exactly the spam detection method mentioned by Dutchman should be implemented?

[admin]/admin/server/admin_chat.lua
```diff
@@ -0,0 +1,158 @@
...
+
+addEventHandler("onPlayerChat", root,
+    function(messageContent, messageType)
+        if messageType ~= 0 then
+            return
+        end
+
+        cancelEvent()
+
+        if utf8.len(messageContent) < 1 then
```
Comment (jlillis) [inline [admin]/admin/server/admin_chat.lua:54]: I think we can skip sending an error about empty messages.

[admin]/admin/server/admin_chat.lua
```diff
@@ -0,0 +1,158 @@
...
+)
+
+addEventHandler("onPlayerQuit", root,
+    function()
+        chatMessages[source] = nil
+        chatTimeouts[source] = nil
+    end
+)
+
+addEventHandler("onPlayerChat", root,
```
Comment (jlillis) [inline [admin]/admin/server/admin_chat.lua:46]: Use a named function for this handler so it's easier to see in `ipb`.

[admin]/admin2/meta.xml
```diff
@@ -67,6 +67,13 @@
 	<script src="client/widgets/admin_warp.lua" type="client" cache="false"/>
 	<script src="client/widgets/admin_report.lua" type="client" cache="false"/>
 	<script src="client/widgets/admin_weapon.lua" type="client" cache="false"/>
+
+	<script src="shared/utils.lua" type="shared" cache="false"/>
+
+	<html src="http/screenshots.htm"/>
```
Comment (jlillis) [inline [admin]/admin2/meta.xml:73]: I think changes on lines 73-76 don't belong in this PR.

jlillis (commented):
Great work overall, just a first round of notes.

0xHexadecimal:
> Strongly disagree. Anti-spam is a basic security task and the admin panel is the best place to do it. It also happens to be a convenient place to add the logic needed to handle player nametag colors and color codes in chat rather than having it re-implemented multiple times across other resources. If server owners already have their own systems in place, they can turn these features off.

I am not against having admin filter chat messages that may be considered harmful. But having a chat-system which cancels the original chat just to display coloured name tags which are set in a different resource, just seems out of place for an admin system. At that point why not just set the random coloured name tags in admin? Along side attaching coloured blips? Why are those not in admin but having a coloured names chatbox is?

PlatinMTA:
The filtering should be doing inside the chat resource and not the admin resource. The admin panel does not handle the chat. If you want the admin panel to have an option to enable or disable the chat maybe add a setting to the meta.xml for the filtering or create some exports.

Nico8340:
> The filtering should be done inside the chat resource and not the admin resource (unless you want to have all the chat in the admin panel). The admin panel does not handle the chat. If you want the admin panel to have an option to enable or disable the chat maybe add a setting to the meta.xml for the filtering or create some exports.

The current implementation handles the chat in the admin resources. (admin and admin2)

[admin]/admin/conf/chat.xml
Comment (Nico8340) [reply [admin]/admin/conf/chat.xml]: Done.

[admin]/admin/server/admin_chat.lua
Comment (Nico8340) [reply [admin]/admin/server/admin_chat.lua]: Done.

[admin]/admin/server/admin_chat.lua
Comment (Nico8340) [reply [admin]/admin/server/admin_chat.lua]: Done.

[admin]/admin2/meta.xml
Comment (Nico8340) [reply [admin]/admin2/meta.xml:73]: Removed.

Nico8340:
I fixed the errors in the review and added the option to completely disable the chat handler. If anyone has an idea for implementing the spam filter mentioned by Dutchman, feel free to share it.

[admin]/admin2/server/admin_chat.lua
```diff
@@ -16,6 +16,10 @@ addEventHandler("onResourceStart", resourceRoot,
         for k, v in pairs(getElementsByType("player")) do
             chatMessages[v] = {}
         end
+
+        if getConfig("isChatEnabled") then
```
Comment (Dezash) [inline [admin]/admin2/server/admin_chat.lua:20]: If IsChatEnabled is set to false, the chatMessages table will no longer be populated causing chat history in player reports to be empty, which is a breaking change.

[admin]/admin/server/admin_server.lua
```diff
@@ -1452,30 +1449,36 @@ addEventHandler ( "aServer", root, function ( action, data, data2 )
 	return false
 end )
 
-addEventHandler ( "onPlayerChat", root, function ( message )
-	local size = #chatHistory[source]
-	if ( size == g_Prefs.maxchatmsgs ) then
```
Comment (Dezash) [inline [admin]/admin/server/admin_server.lua:1457]: The new chat handler does not have the maxchatmsgs setting implemented. If it is deemed unnecessary then you should also remove it from meta.xml and admin_serverprefs

[admin]/admin2/server/admin_chat.lua
Comment (Nico8340) [reply [admin]/admin2/server/admin_chat.lua:20]: Which is the better solution in your opinion? If we create a separate event handler for logging messages, or do we always create the event handler, but if the "isChatEnabled" config is not true, then only the logging happens?

Comment (Dezash) [reply [admin]/admin2/server/admin_chat.lua:20]: Always having the handler would be better in my opinion. You could also move the settings to admin_serverprefs which would handle the caching and allow altering these settings without restarting the resource

Dutchman101:
> > It would be nice to use this opportunity for blocking common spam characters (like Bismillah).. there is a way to select by charset to stop weird characters like that, if you don't know how i will post it soon
> 
> How do you imagine the implementation? Ban certain keywords or limit the message to a certain character set? (a-Z, 0-9, etc)

@Nico8340 

This will eliminate a bunch of spam characters:
```lua
function removeChatSpam(msg)
	local newMsg = ""
	for i = 1, utfLen(msg) do
		local char = utfCode(utfSub(msg,i,i))
		if(not (char >= 768 and char <= 879)) then
			newMsg = newMsg .. utfSub(msg,i,i)
		end
	end
	return newMsg
end
```

[admin]/admin/server/admin_server.lua
Comment (Nico8340) [reply [admin]/admin/server/admin_server.lua:1455]: Still in use, by admin_server.lua.

Dutchman101:
There's a conflict with d8834a8 (i dont know why it's not been detected)

Nico8340:
> There's a conflict with [d8834a8](https://github.com/multitheftauto/mtasa-resources/commit/d8834a86c60490d7919f8e4b05ac70946f420279) (i dont know why it's not been detected)
I see that the whole Pull Request is bugged, but since there are still a lot of unnecessary commits and I messed up the whole branching, I'm closing this PR and opening a new PR soon.

Dutchman101:
How about some advanced anti-spam features, like:

- Configurable limits with action, like: mute, kick, or ban player for x duration if they repeatedly get flagged for spamming. Mute is the first line defense, to store serials of repeat offenders for persisting spamming rap sheet, an SQLite .db can be used. The more times they get a lighter automatic punishment (mute, kick), the sooner it will escalate to a more severe punishment, Mute > kick > 30 mins ban > longer bans as defined in meta.xml. In this case, also add a toggle for automatic .db clearing (choice between interval or to do it on server launch)

**Measures against coordinated spamming, like:**
Multiple players are sending the same or similar message (not a single source, but it'll be big) or different ones; measures like anti-repeat or chat limits _per player_ won't be effective
* If chat activity exceeds a certain amount (specified in meta.xml) of messages per the polling window (specified in meta.xml, default: 10 secs), an abnormality has started and the system will offer a meta.xml toggle-based resolution: slowmode (limit chat messages globally, from anyone.. and return "You need to wait, slowmode has been activated due to possible spam" to anyone that tries to talk while it's still active

So a spamming incident can be resolved without the presence of a server admin. If one is present when slowmode is activated, they should be able to use a command to lift it if they go to manually resolve it (like: ban the coordinated spammers)

Nico8340:
> How about some advanced anti-spam features, like:
> 
> * Configurable limits with action, like: mute, kick, or ban player for x duration if they repeatedly get flagged for spamming. Mute is the first line defense, to store serials of repeat offenders for persisting spamming rap sheet, an SQLite .db can be used. The more times they get a lighter automatic punishment (mute, kick), the sooner it will escalate to a more severe punishment, Mute > kick > 30 mins ban > longer bans as defined in meta.xml. In this case, also add a toggle for automatic .db clearing (choice between interval or to do it on server launch)
> 
> **Measures against coordinated spamming, like:** Multiple players are sending the same or similar message (not a single source, but it'll be big) or different ones; measures like anti-repeat or chat limits _per player_ won't be effective
> 
> * If chat activity exceeds a certain amount (specified in meta.xml) of messages per the polling window (specified in meta.xml, default: 10 secs), an abnormality has started and the system will offer a meta.xml toggle-based resolution: slowmode (limit chat messages globally, from anyone.. and return "You need to wait, slowmode has been activated due to possible spam" to anyone that tries to talk while it's still active
> 
> Server owners can tweak the # of messages/polling interval based on what's normal for their server. For example, if it's a gamemode some people can get hyped up and send a burst of happy messages, like end-of-round, they would need to make the polling interval longer. If the server gets more popular (leading to increased chat activity), they would need a higher message count tolerance.
> 
> So a spamming incident can be resolved without the presence of a server admin. If one is present when slowmode is activated, they should be able to use a command to lift it if they go to manually resolve it (like: ban the coordinated spammers)
> 
> @Nico8340

It's fine by me, please copy this message to the new PR when i open it, and continue any future conversation there.

PR #464, opened 2024-01-28
Improved Msgbox

Nico8340:
This PR removes the old msgbox resource and creates a new one that provides a much simpler concept and looks better.
Available types: info, question, warning, error
Available buttons: ok, okcancel, retrycancel, yesno, yesnocancel
Képernyőkép 2024-01-28 012351


Nico8340:
A set of pictures of the new look in action (default resources)
image
image
image

Fernando-A-Rocha:
Good job! I would just add more spacing between the title and message, as well as the icon on the left @Nico8340

Nico8340:
> Good job! I would just add more spacing between the title and message, as well as the icon on the left @Nico8340

Should there be more space vertically or horizontally?
@Fernando-A-Rocha

Fernando-A-Rocha:
> > Good job! I would just add more spacing between the title and message, as well as the icon on the left @Nico8340
> 
> Should there be more space vertically or horizontally?
> @Fernando-A-Rocha 

On the y axis between the 2 text labels, and on the x axis between the image and the texts, imo @Nico8340

Nico8340:
> > > Good job! I would just add more spacing between the title and message, as well as the icon on the left @Nico8340
> > 
> > 
> > Should there be more space vertically or horizontally?
> > @Fernando-A-Rocha
> 
> On the y axis between the 2 text labels, and on the x axis between the image and the texts, imo @Nico8340

Does they look better now? 😄 
image

Fernando-A-Rocha:
> > On the y axis between the 2 text labels, and on the x axis between the image and the texts, imo @Nico8340
> 
> Does they look better now? 😄 
> @Fernando-A-Rocha 
> image
> 

I think it looks fine now

Fernando-A-Rocha:
There is no way to handle when a player clicked a specific button like Ok, Cancel? Seems like all they do is close the window

Nico8340:
> There is no way to handle when a player clicked a specific button like Ok, Cancel? Seems like all they do is close the window

Of course, it is possible to handle the buttons, the function returns three variables that contain the buttons, and the onClientGUIClick event can be used to handle the responses. They only closes so that the scripts do not have to handle this separately.

Nico8340:
By the way, I would like to use this system for the new chat implementation as well, in case of mute, notify the player to end this activity or he will face a bigger sanction.

[editor]/editor_gui/client/me_gui.lua
```diff
@@ -370,9 +370,9 @@ function playSoundFrontEnd ( sound )
 	return returnValue
 end
 
-function guiShowMessageBox ( message, boxType, title, forced, button1, button2, button3 )
+function guiShowMessageBox ( title, msg, type, buttons )
 	playSoundFrontEnd ( 5 )
```
Comment (jlillis) [inline [editor]/editor_gui/client/me_gui.lua:374]: Perhaps `msgbox` should make the calls to `playSoundFrontEnd` so behavior is consistent regardless of the calling resource?

[gameplay]/msgbox/meta.xml
```diff
@@ -0,0 +1,12 @@
+<meta>
+    <info author="Nico834" version="1.0" name="MsgBox" description="Create and manage message boxes." type="script" />
+    <min_mta_version server="1.6.0" client="1.6.0" />
+
+    <script src="sourceC.lua" type="client" cache="false" />
```
Comment (jlillis) [inline [gameplay]/msgbox/meta.xml:5]: This file isn't security-critical (imo) so I think this should be cacheable.

[gameplay]/msgbox/meta.xml
```diff
@@ -0,0 +1,12 @@
...
+    <export function="createMessageBox" type="client" />
+
```
Comment (jlillis) [inline [gameplay]/msgbox/meta.xml:7]: For the rare and unlikely case other resources by others (outside this repo) used the original `msgbox` exports you should consider re-adding the old syntax as a backwards-compatible alternative.

[gameplay]/msgbox/sourceC.lua
```diff
@@ -0,0 +1,136 @@
...
+    ["yesnocancel"] = {
+        [1] = "Yes",
+        [2] = "No",
+        [3] = "Cancel"
+    }
+}
+
+function createMessageBox(boxTitle, boxMessage, boxType, boxButtons)
+    if not boxTitle or type(boxTitle) ~= "string" then
+        outputDebugString("Bad argument @ 'createMessageBox' [Expected string at argument 1, got " .. type(boxTitle) .. "]", 2)
```
Comment (jlillis) [inline [gameplay]/msgbox/sourceC.lua:48]: For these checks consider using `error(message, 2)`. This will invoke an error in the calling resource rather than within `msgbox`, making debugging easier.

[gameplay]/msgbox/sourceC.lua
```diff
@@ -0,0 +1,136 @@
...
+        outputDebugString("Bad argument @ 'createMessageBox' [Unknown messageBoxType at argument 3", 2)
+        return
+    end
+
+    if not messageBoxButtons[boxButtons] then
+        outputDebugString("Bad argument @ 'createMessageBox' [Unknown messageBoxButtons at argument 4", 2)
+        return
+    end
+
+    local boxWidth = 350
```
Comment (jlillis) [inline [gameplay]/msgbox/sourceC.lua:76]: I think these static values could be defined in a table at the top of script rather than in this function.

jlillis (changes requested):
Looks good overall, just a few minor notes. This implementation will work just fine in `editor_gui` and similar resources, but the `admin` and `admin2` resources have their own unique implementations that make use of coroutines that can't be easily replaced by this resource.

[gameplay]/msgbox/sourceC.lua
Comment (Nico8340) [reply [gameplay]/msgbox/sourceC.lua]: Done

[editor]/editor_gui/client/me_gui.lua
Comment (Nico8340) [reply [editor]/editor_gui/client/me_gui.lua:374]: Done

[gameplay]/msgbox/sourceC.lua
Comment (Nico8340) [reply [gameplay]/msgbox/sourceC.lua:76]: Done

[gameplay]/msgbox/meta.xml
Comment (Nico8340) [reply [gameplay]/msgbox/meta.xml]: Done

[gameplay]/msgbox/meta.xml
Comment (Nico8340) [reply [gameplay]/msgbox/meta.xml:8]: Done

Nico8340:
The changes have been made, but for some reason GitHub says they haven't.

Nico8340:
I'm closing this pull request because I have a much better idea for the implementation that is completely different from this one.

PR #466, opened 2024-02-02
Update assault.lua

jvstns:
- Change source to client, as source can be manipulated
- Fully remove the player data from the **assaultClientScriptQueue** table when the player quits.


Fernando-A-Rocha (commented):
LGTM!

PR #468, opened 2024-02-10
admin2: Replacing checkboxes with a gridlist for glitches and special world properties

FileEX:
Recently, new special world properties have appeared, which means that there is no space for new properties (next checkboxes) in the admin panel. I believe that replacing checkboxes with a gridlist is the best solution.

Old list (without enough space for two new properties)
image

New solution:
image

To enable or disable a glitch or a special world property, just double-click on a row in the gridlist.

I added two new properties ``burnedflippedcars`` and ``fireballdestruct``


jlillis:
Looks great. Do you think this could be easily back-ported to `admin`?

FileEX:
I think we should also back-ported it to admin. Especially since it is not known yet when admin will be replaced by admin2.

PR #471, opened 2024-02-21
fix issue where defaultstats did not set player stats correctly (fixes #131)

### LINKED ISSUE #131: Maxed out skills resets while testing map
defaultstats

**To Reproduce**
1. Start Freeroam, then defaultstats.
2. Spawn any bicycle, hold jump button and release to make high jump.
3. Start Map Editor and press F5.

Now repeat the same steps of number 2. 

**Expected behaviour**
Everytime Freeroam starts after press Full Test your maxed out skills resets because defaultstats automatically stop so you can't do bicycle high jump or shoot while walking, for example.

MittellBuurman:
- changed resourceRoot to root in defaultstats.lua
- added conditions in applyStatsForEveryone to only apply stats when resource `type` is `gamemode` or `map`; or the resource itself.

### DIFF
--- a/[gameplay]/defaultstats/defaultstats.lua
+++ b/[gameplay]/defaultstats/defaultstats.lua
@@ -26,12 +26,18 @@ local function applyStatsForSource()
 end
 addEventHandler("onPlayerJoin", root, applyStatsForSource)
 
-local function applyStatsForEveryone()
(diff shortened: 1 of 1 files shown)


Fernando-A-Rocha (commented):
All good. Someone with access pls approve this PR because I can't

PR #472, opened 2024-04-19
Add option to disable default report system.

CArg22:
Adds option "reportsEnabled" that can be used to disable or enabled default reports. Often servers are implementing this kind of system by themself so to do not duplicate functionality they can easily disable default report system.

### DIFF
--- a/[admin]/admin/client/gui/admin_report.lua
+++ b/[admin]/admin/client/gui/admin_report.lua
@@ -4,7 +4,7 @@
 *
 *	gui\admin_report.lua
 *
-*	Original File by lil_Toady
+*	Original File by lil_Toady|CArg22
 *
 **************************************]]
 
@@ -58,8 +58,6 @@ function aReport ( )
 	guiBringToFront ( aReportForm )
 	showCursor ( true )
 end
-addCommandHandler ( "report", aReport )
-
 
 function aReportClose ( )
 	if ( aReportForm ) then
@@ -171,3 +169,10 @@ function aClientReportClick ( button )
 		end
 	end
 end
+
+addEvent("aClientReports",true)
+addEventHandler("aClientReports",resourceRoot,
+	function( )
+		aReport ( )
+	end
+)
--- a/[admin]/admin/meta.xml
+++ b/[admin]/admin/meta.xml
@@ -18,6 +18,7 @@
   <script src="server/admin_ACL.lua" />
   <script src="server/admin_settings.lua" />
   <script src="server/admin_screenshot.lua" />
+  <script src="server/admin_report.lua" />
   <script src="server/http/admin_http.lua" />
   <script src="client/admin_clientprefs.lua" type="client" cache="false" />
   <script src="client/admin_client.lua" type="client" cache="false" />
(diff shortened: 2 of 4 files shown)


FileEX:
The idea is good because now you have to manually fiddle with the file to disable the default reports. Many servers have their own reporting systems, which makes reports from the admin panel unnecessary and the ability to turn them off is good.

If the code will be revised and tested it can also be forward-ported to admin2

[admin]/admin/client/gui/admin_report.lua
```diff
@@ -4,7 +4,7 @@
 *
 *	gui\admin_report.lua
 *
-*	Original File by lil_Toady
+*	Original File by lil_Toady|CArg22
```
Comment (Nico8340) [inline [admin]/admin/client/gui/admin_report.lua:7]: If the original code was not created by you, I don't think you should write your name in it, but let's hear other people's opinions about it.

Comment (Lpsd) [reply [admin]/admin/client/gui/admin_report.lua:7]: Yeah please revert this

[admin]/admin/server/admin_report.lua
```diff
@@ -0,0 +1,15 @@
+--[[**********************************
+*
+*	Multi Theft Auto - Admin Panel
+*
+*	admin_reports.lua
+*
+*	Original File by CArg22
+*
+**************************************]]
```
Comment (Lpsd) [inline [admin]/admin/server/admin_report.lua:9]: Remove this 2006 block comment, it's bigger than the code itself.

Lpsd (changes requested):

[admin]/admin/client/gui/admin_report.lua
```diff
@@ -58,8 +58,6 @@ function aReport ( )
 	guiBringToFront ( aReportForm )
 	showCursor ( true )
 end
-addCommandHandler ( "report", aReport )
```
Comment (jlillis) [inline [admin]/admin/client/gui/admin_report.lua:61]: Removing this command handler removes functionality that some users may want to use. I would put this back.

[admin]/admin/client/gui/admin_report.lua
```diff
@@ -171,3 +169,10 @@ function aClientReportClick ( button )
 		end
 	end
 end
+
+addEvent("aClientReports",true)
+addEventHandler("aClientReports",resourceRoot,
+	function( )
+		aReport ( )
+	end
+)
```
Comment (jlillis) [inline [admin]/admin/client/gui/admin_report.lua:178]: This could be shortened to `addEventHandler("onClientReports, resourceRoot, aReport)`. No need for an anonymous function here.

[admin]/admin/meta.xml
```diff
@@ -176,5 +177,10 @@
           				desc="The command name for admin chat."
           				/>
 
+		<setting name="*reportsEnabled" value="true"
+					friendlyname="Whatever reports are enabled"
+					desc="Set to true to enable default reporting system"
+					/>
```
Comment (jlillis) [inline [admin]/admin/meta.xml:183]: Some grammar improvements could be made here:
`friendlyname="Enable default reporting system"
desc="When enabled, players will be able to submit reports to admins via the /report command."`

jlillis (changes requested):
I like the idea but have a few issues with this PR as is:
1 - There isn't a clear indication to other admins/moderators that the reporting system is disabled anywhere in the UI.
2 - If this change is made for `admin`, it should also be made for `admin2`.

Nico8340:
Any updates on this?

Dutchman101:
Superseded by #501

PR #474, opened 2024-04-30
playerblips: Fixed that the resource "playercolors" should be activated for teams

T-MaxWiese-T:
The reason for this pull request is because the resource did not work the way I wanted it to. Only after I had a closer look at the code I realized that you can configure the playerblips resource in the `meta.xml`. Because I wanted the playerblips resource to take over the team colors that I have in my code, but that didn't happen. And since the server log said to activate the playercolors resource, I did it, but it didn't help. Only after I deactivated the playercolors resource and set "`use_team_colors`" to "`true`" did it work as expected. 
Anyway, the playercolors resource only makes sense if you want to have random player nametag colors, which makes no sense for teams, since teams can only have one color.

### DIFF
--- a/[gameplay]/playerblips/server/main.lua
+++ b/[gameplay]/playerblips/server/main.lua
@@ -13,7 +13,7 @@ local function resourceStart()
 	end
 
 	local playercolorsResource = getResourceFromName("playercolors")
-	if playercolorsResource and getResourceState(playercolorsResource) ~= "running" then
+	if playercolorsResource and getResourceState(playercolorsResource) ~= "running" and not useTeams then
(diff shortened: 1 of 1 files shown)


jlillis:
> I realized that you can configure the playerblips resource in the `meta.xml`.

The intended use case is for admins to change the settings via the admin panel, but you can also do it directly via the meta.xml. This PR brings up another issue I just noticed - the resource has no event handlers for [`onSettingChange`](https://wiki.multitheftauto.com/wiki/OnSettingChange) so it relies on admins knowing to restart the resource after changing a setting (a step that isn't documented anywhere). I'll open an issue to fix that separate. Your PR does improve functionality when using team colors so I'll merge it now. Congratulations on your first contribution here!

PR #483, opened 2024-05-10
Making the admin window focused

Proxy-99:
making the window focused  when pressing P to open


[admin]/admin2/client/main/admin.lua
```diff
@@ -34,6 +34,7 @@ addEventHandler(
             aAdminMain.Close(false)
         else
             aAdminMain.Open()
+		    guiFocus (aAdminMain.Form)
```
Comment (Nico8340) [inline [admin]/admin2/client/main/admin.lua:37]: There is an unnecessary indent here, do you have any reason for it?

Nico8340 (commented):
Looks good to me 🚀

PR #490, opened 2024-05-13
admin: Resource settings button is only displayed if there are settings

### LINKED ISSUE #489: admin: Resource settings button should only be displayed if there are settings
Currently, when you click on a resource in the resource list, you cannot see directly whether this resource has settings. 
If the button with the resource settings would only be displayed if there are settings, it would save a lot of time and work, especially when checking multiple resources.

Resource settings button should only be displayed if there are settings.

None.

**Additional context**
I have looked at the admin2 resource and there the resource settings are displayed directly, which I think is definitely better.
I was also thinking that you could perhaps hide all other buttons such as “Start” or “Stop” as long as no resource has been selected, which would make the pointless error message that no resource has been selected obsolete.

T-MaxWiese-T:
This pull request refers to this feature request (closes #489) about the admin resource where the resource settings button should only be displayed if there are settings.
It was difficult to make this change. It felt like I had to study half the code to get an overview.


Fernando-A-Rocha:
I think the same change should be made to the new admin2 resource because admin "1" will be discontinued afaik

T-MaxWiese-T:
> I think the same change should be made to the new admin2 resource because admin "1" will be discontinued afaik

I assume that you have not tried the latest admin2 resource version. But I have already written the following about the admin2 resource in my issue (#489):
> **Additional context** I have looked at the admin2 resource and there the resource settings are displayed directly, which I think is definitely better.

However, I have considered that instead of an empty grid list if a resource has no settings, it should say “None”, for example, as well as if no description is given, for example.
Because a good GUI gives you information. If the user sees an empty grid list, he might think that there is an error during loading. It is also clearer if only what is needed is displayed.

Dutchman101:
Thanks!

PR #491, opened 2024-05-17
Refactor of resources meta.xml

Fernando-A-Rocha:
This PR introduces the following changes:

- author tag added in resources info that didn't have it (`MTA contributors (github.com/multitheftauto/mtasa-resources)`)
- version tag added in resources info that didn't have it (`1.0.0`)
- unnecessary `<?xml` line removed from all .map and meta.xml files that had it
- unused `mod="deathmatch"` tag removed from all .map files`

Feel free to give your opinion.

To me, the most important change is adding `version` tags to all resources, so version control standards can be enforced like they are in community.mtasa.com. Example of good practice: https://community.multitheftauto.com/index.php?p=resources&s=details&id=3716


Fernando-A-Rocha:
> Does it mean that from now on if a resource is changed for a future release the version number has to be increased?

I believe so. This is the responsibility of us, contributors and maintainers of this repository. I don't see why a resource version shouldn't be bumped when it's changed, despite the fact that it's not that useful for now. Maybe in the future there can be a resource update checker.

T-MaxWiese-T:
> > Does it mean that from now on if a resource is changed for a future release the version number has to be increased?
> 
> I believe so. This is the responsibility of us, contributors and maintainers of this repository. I don't see why a resource version shouldn't be bumped when it's changed, despite the fact that it's not that useful for now. Maybe in the future there can be a resource update checker.

It is already possible to update resources via the server console. 
The admin panel should actually also have this option. There should be a new lua function for this: 
`bool upgradeResources ( [ bool upgradeAll = true, resource targetResource = nil ] )`
And a function that returns the resource names that need to be updated and the latest version:
`table getUpgradeResources ( [ bool allResources = true, resource targetResource = nil ] )`

PR #494, opened 2024-05-26
Fixed nil index error and removed invalid characters causing syntax errors

rad3sh:
- Moved the initialization of 'aLogMessages[type][action]' to after the definition of 'action'.
- Removed non-space invalid characters that were causing syntax errors.


Dutchman101:
Thanks

PR #495, opened 2024-05-29
[Editor] Various fixes for local spawned or invalid elements.

chris1384:
Some resources create their own elements but are not synced to the server. Doing an action to that certain element can result in looping errors and editor suspending.
This commit fixes all of those.


Dutchman101:
Thanks, @chris1384

PR #500, opened 2024-06-14
internet radio: fix few things

srslyyyy:
- Better naming for table with active radios
- Change event sources
- Save colors to variable

### DIFF
--- a/[gameplay]/internetradio/client.lua
+++ b/[gameplay]/internetradio/client.lua
@@ -1,6 +1,6 @@
-local resX, resY = guiGetScreenSize()
-local tbl = {}
-
+local radioSpeakers = {}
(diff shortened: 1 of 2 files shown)


Fernando-A-Rocha:
@srslyyyy Is it not a security concern the fact that player serials are being sent to every client for them to store in a table?

Dutchman101:
> @srslyyyy Is it not a security concern the fact that player serials are being sent to every client for them to store in a table?

He's planning to refactor it even more, but this is a good start. Btw, serial faking has gotten a definite fix over a month ago, yet the practise needs to change; it's not brand new btw, i pulled this popular resource from my server and only did a few tweaks.

PR #501, opened 2024-06-15
Allow disabling/enabling default reporting system

Viude:
This allows disabling/enabling default reporting system. This is my first pull request, so I hope everything is in order.


Nico8340 (commented):
Good job 👍

PR #502, opened 2024-06-18
Update clientcheckban setting to ban serial instead of IP

Viude:


[admin]/admin/meta.xml
```diff
@@ -142,20 +142,39 @@
 					desc="Detect fake admin packets. 0-No checks  1-Some checks  2-All checks."
 					/>
 
-		<setting name="*clientcheckban" value="false"
```
Comment (T-MaxWiese-T) [inline [admin]/admin/meta.xml:145]: I would prefer to move the new settings here.

[admin]/admin/server/admin_serverprefs.lua
```diff
@@ -18,7 +18,10 @@ function cachePrefs()
 	g_Prefs.mutedurations	= getString('mutedurations','60,120,300,600,0')
```
Comment (T-MaxWiese-T) [inline [admin]/admin/server/admin_serverprefs.lua:18]: In this function, the unnecessary spaces in front of the "=" symbols could be removed.

[admin]/admin/meta.xml
```diff
@@ -142,20 +142,39 @@
 					desc="Detect fake admin packets. 0-No checks  1-Some checks  2-All checks."
```
Comment (T-MaxWiese-T) [inline [admin]/admin/meta.xml:142]: I would move the setting "*securitylevel" to the group "_fake packets" and add it because it belongs together.

[admin]/admin/meta.xml
```diff
@@ -137,7 +137,7 @@
 
 		<setting name="*securitylevel" value="2"
```
Comment (T-MaxWiese-T) [inline [admin]/admin/meta.xml:138]: Move the setting down to the other settings in the same group.

Nico8340 (commented):
Good job 👍

PR #506, opened 2024-06-28
admin: Fixed that double clicking on a resource without setting opened the GUI settings window

T-MaxWiese-T:
This pull request ensures that the GUI settings window is no longer opened if you double-click on a resource name in the resource list and it has no settings. 
In this pull request https://github.com/multitheftauto/mtasa-resources/pull/490 with the change that the settings button should only be displayed if there are settings, I have not considered this.


Nico8340 (commented):
LGTM

Dutchman101:
Thanks!

PR #507, opened 2024-06-30
internetradio: refactor

srslyyyy:
- Refactored code
- More config options
- Data gets stored under player element now
- .json as setting file
- Settings are verified, if something doesn't match, value gets reverted to defaultsTo, unused settings are automatically cleaned
- Introduced some optimisations (probably there would be few extra)
- Added delays to prevent player from abusing triggers
- Allow other players speakers doesn't require reconnect anymore
- Used patterns for defining .lua scripts in meta.xml


srslyyyy:
Ready for testing

[gameplay]/internetradio/config/CRadioConfig.lua
```diff
@@ -0,0 +1,23 @@
...
+-- ## Authors: MTA contributors			##
+-- ## Version: 1.0						##
+-- #######################################
+
+RADIO_TRACK_SCALE = 1
+RADIO_TRACK_FONT = "default-bold"
+RADIO_TRACK_COLOR = tocolor(150, 50, 150, 255)
+RADIO_TRACK_BACKGROUND_COLOR = tocolor(0, 0, 0, 255)
+
+RADIO_TOGGLE_KEY = "F9"
```
Comment (T-MaxWiese-T) [inline [gameplay]/internetradio/config/CRadioConfig.lua:12]: The "F9" key is already assigned to the resource "helpmanager" https://github.com/multitheftauto/mtasa-resources/pull/498/files .

[gameplay]/internetradio/config/CRadioConfig.lua
```diff
@@ -0,0 +1,23 @@
+-- #######################################
+-- ## Project: Internet radio			##
+-- ## Authors: MTA contributors			##
+-- ## Version: 1.0						##
```
Comment (T-MaxWiese-T) [inline [gameplay]/internetradio/config/CRadioConfig.lua:4]: Why are the "`#`" symbols arranged so strangely in the files?

Comment (srslyyyy) [reply [gameplay]/internetradio/config/CRadioConfig.lua:4]: It's GH issue, try downloading it directly from branch, they would appear as they should.

[gameplay]/internetradio/config/CRadioConfig.lua
Comment (srslyyyy) [reply [gameplay]/internetradio/config/CRadioConfig.lua:12]: What do you propose?

Comment (T-MaxWiese-T) [reply [gameplay]/internetradio/config/CRadioConfig.lua:12]: > Which key do you propose?

The key I used in this pull request: https://github.com/multitheftauto/mtasa-resources/pull/498/files .
I had actually worked on the option to change the key via the resource settings.

[gameplay]/internetradio/config/CRadioConfig.lua
Comment (T-MaxWiese-T) [reply [gameplay]/internetradio/config/CRadioConfig.lua:4]: > It's GH issue, try downloading it directly from branch, they would appear as they should.

Yes.

[gameplay]/internetradio/config/CRadioConfig.lua
Comment (Fernando-A-Rocha) [reply [gameplay]/internetradio/config/CRadioConfig.lua:12]: Changing via resource setting is a great idea.

Fernando-A-Rocha:
I noticed that all config in this resource is done in Lua. Due to the amount of modifiable variables and their complexity (table of radio stations, rgb colors, etc), I'm okay with keeping them as is, and not making meta.xml settings for this resource. Unless anybody thinks it would be easy to put all this as xml settings...

T-MaxWiese-T:
> I noticed that all config in this resource is done in Lua. Due to the amount of modifiable variables and their complexity (table of radio stations, rgb colors, etc), I'm okay with keeping them as is, and not making meta.xml settings for this resource. Unless anybody thinks it would be easy to put all this as xml settings...

I think it would make sense for most variables. And with the table you could put each radio station in a group, which means you could also enter your own radio name and URL. There is also the possibility to work with tables in the settings, but this is not very attractive due to the small width of the text field for changes and the display. In admin2 you could consider making the text field larger so that you would have a much better overview and you would have to simplify the display and thus translate it. But the bottom line is that I find this rather impractical. It's still easiest if you have individual variables that you can change.

Fernando-A-Rocha:
You say that it's viable to have all settings in Meta.xml, I'm interested in seeing how that can be implemented 😅 what do you think? @srslyyyy

Fernando-A-Rocha:
I agree tbh, it is unclear how to make this resource more easily configurable for now. Lua vars is fine. Feel free to merge, imo.

jlillis:
I don't think we need all the possible configuration options available in via admin settings. The only one I can think of would be nice is the setting that allows/disallows players to place speaker objects.

jlillis (approved):

Nico8340 (commented):
I just tested the refactored resource.
It's perfect along with the code 🎉

Dutchman101 (approved):

PR #511, opened 2024-07-04
Maintenance map editor

IIYAMA12:
Fixes:
- An issue with [unloading the edf](https://github.com/IIYAMA12/mtasa-resources/blob/2ac28d7af47153e32f7f99bb2ddfe648fb6a700e/%5Beditor%5D/editor_main/server/EDFhandler.lua#L44) (which created a lot of lag when stopping the editor resource)
- Ped armor related (`get` had to be used instead of `set`)
- colpath.lua: Not enough checks for nil values
- Wrapper func setPlayerBlipSize was not using the size parameter. Not that it really matters, afaik this function was not being used anyway.
- Making sure that the editor stops edf resources, which didn't do anything before. Although I am not sure how this functionality actually works.
- A typo in matrix.lua (the file was also used in the resource sirenEdit)
- Allow the element lock functionality to be used on edf elements
- Distinct `Locked vehicle` from `Locked selection`. I wanted to give it a better name, but the name `unselectable` didn't fit in the GUI, so I used `locked-s` instead (When hovering: `Locked selection`).


Nico8340 (commented):
Awesome 💯

[editor]/edf/edf.lua
```diff
@@ -620,7 +620,11 @@ function edfCreateElement(elementType, creatorClient, fromResource, parametersTa
 			elseif dataField == "rotation" then
 				edfSetElementRotation(newElement, dataValue[1], dataValue[2], dataValue[3], dataValue[4])
 			elseif dataField == "interior" then
-				setElementInterior(newElement, dataValue)
+				if dataValue == -1 then
```
Comment (Xenius97) [inline [editor]/edf/edf.lua:623]: Why not just use math.max?

```lua
setElementInterior(newElement, math.max(dataValue, 0))
```

Comment (IIYAMA12) [reply [editor]/edf/edf.lua:623]: While `math.max` does technically solve the problem, the code itself does not explain (without comment) that there is an option -1 available. 
It also hides the warning if an -2, -3, -4... value slips in, which can be a good thing but in my opinion it is an early warning that the map file parser might has been broken.

Comment (Xenius97) [reply [editor]/edf/edf.lua:623]: setElementInterior has no minus values, only setElementDimension does.

Comment (IIYAMA12) [reply [editor]/edf/edf.lua:623]: > interior: The interior you want to set the element to. Valid values are 0 to 255.
Yes, but the input range is -1 to 255. And not 0 to 255. The use of math.max does not explain that... 
Having at least a little bit of self-explaining code couldn't hurt right?

Comment (Xenius97) [reply [editor]/edf/edf.lua:623]: For setElementInterior, negative values still aren't allowed, only 0-255. I didn't make this up myself, you overcomplicated it, but as I said, it will work.

Comment (IIYAMA12) [reply [editor]/edf/edf.lua:623]: > For setElementInterior, negative values still aren't allowed, only 0-255
I know that, but I was talking about the map editor code, which uses -1 to 255 respectively. And not the input value for setElementInterior.
But I am going to close the conversation. I added the if statement for a specific reason and not just for a value restrain of 0-255.

PR #516, opened 2024-07-13
Freeroam security patch

IIYAMA12:
Currently the Freeroam Remote Player Call Functions it's parameters can be freely manipulated by any client that uses code injections.

What does this pull request includes?
- Adding parameter validation to the Remote Player Call Functions. (main patch)
- Some other tweaks to future proof some of the Remote Player Call Functions. 
   - Note some of my changes were already patched at: https://github.com/multitheftauto/mtasa-resources/blob/c96bdd5297cf180f947596c1eded8929b4982e6c/%5Bgameplay%5D/freeroam/fr_server.lua#L600 but hardcoded is always better.


[gameplay]/freeroam/fr_server.lua
```diff
@@ -137,7 +134,8 @@ function joinHandler(player)
 		outputChatBox('Press F1 to show/hide controls', player, 0, 255, 0)
 	end
 end
-addEventHandler('onPlayerJoin', root, joinHandler)
+
+addEventHandler('onPlayerJoin', root, function() joinHandler(source) end)
```
Comment (T-MaxWiese-T) [inline [gameplay]/freeroam/fr_server.lua:138]: Why did you add the anonymous function? What sense does it make?

[gameplay]/freeroam/meta.xml
```diff
@@ -13,6 +13,7 @@
 	<include resource="playerblips" />
 
 	<script src="util_server.lua" type="server" />
+	<script src="remote_player_call_validation_s.lua" type="server" />
```
Comment (T-MaxWiese-T) [inline [gameplay]/freeroam/meta.xml:16]: The other files for the server have "`_server`" in their name, so to keep it consistent the newly added file should also have "`_server`" in its name.

Comment (IIYAMA12) [reply [gameplay]/freeroam/meta.xml]: Nice, resolved in 21abca9a0e8fd5ceac62565414188118e18b5400

[gameplay]/freeroam/fr_server.lua
Comment (IIYAMA12) [reply [gameplay]/freeroam/fr_server.lua:138]: It is to get rid of this condition:
https://github.com/multitheftauto/mtasa-resources/blob/c96bdd5297cf180f947596c1eded8929b4982e6c/%5Bgameplay%5D/freeroam/fr_server.lua#L127

While in it's current context it is not exploitable, it does make keep room for future vulnerabilities. 
* The origin of the source variable has to be crystal clear for named functions. When a function are called from multiple places, it's origin becomes less clear.

Comment (T-MaxWiese-T) [reply [gameplay]/freeroam/fr_server.lua:138]: I think that's unlikely. `source` is usually from an event and since it is on the server it cannot come from the client. According to the logic, the function would also be vulnerable as all functions could pass all possible variables to the `player` parameter and the origin is not clear.
I thought your intention was to make the code shorter and you could pass `source` as a parameter but then you would have to customize the code. In that case, however, it is not necessarily immediately obvious what `source` should be.  So it's debatable, but I think it's fine even if I think it's an atypical way.

Comment (IIYAMA12) [reply [gameplay]/freeroam/fr_server.lua:138]: > I think that's unlikely. source is usually from an event and since it is on the server it cannot come from the client. According to the logic, the function would also be vulnerable as all functions could pass all possible variables to the player parameter and the origin is not clear.

That is true, but as I said before it is just a precaution. You want people to be aware that `source` can be exploited.

In the second call is also the source (resourceRoot) set:
https://github.com/multitheftauto/mtasa-resources/blob/c96bdd5297cf180f947596c1eded8929b4982e6c/%5Bgameplay%5D/freeroam/fr_server.lua#L113
There should be (in my opinion) no doubt in which values are used inside of a function.

> So it's debatable,

It surely is! :)

PR #521, opened 2024-07-18
Fix minclientversion checker for MTA 1.6 and on.

YelehaUwU:


Fernando-A-Rocha:
I think mtasa.com should be changed to multitheftauto.com in the URLs

https://github.com/multitheftauto/mtasa-resources/blob/2dddd00d2931bb21c0a3e181913e55389c778bb8/%5Badmin%5D/acpanel/s_settings.lua#L201

https://github.com/multitheftauto/mtasa-resources/blob/2dddd00d2931bb21c0a3e181913e55389c778bb8/%5Badmin%5D/acpanel/s_settings.lua#L235

Also, I was unaware there was an endpoint that returns `acpanel` version. Logically, after updating the `acpanel` resource, the version in meta.xml should be bumped and also updated on that webpage.

Additionally, the `outputChatBox("New version of Anti-Cheat panel is available!")` should be switched to an `outputServerLog` imo.

Fernando-A-Rocha:
I found a bug related to acpanel. The URL displayed in leads to https://nightly.mtasa.com/index.php?p=resources&s=details&id=6242 which seems like it should have been on community.multitheftauto.com to land on the acpanel resource page...

image

Dutchman101:
I will add the other requests with some tweaks in a follow up commit; you didn't enable maintainer changes

PR #528, opened 2024-07-29
basic security resource as a foundation

ffsPLASMA:
As the notes say, this resource aims to be a barebone security foundation for server owners to enhance and adapt on for their personal gamemodes. As recent times showed, security needs to be highlighted more and this resource should give a base start point for developers to work on. It only covers the most basic events and logs them.

### DIFF
--- a/[admin]/security/logging.lua
+++ b/[admin]/security/logging.lua
@@ -16,7 +16,6 @@ function logViolation(uPlayer, strMessage)
 	outputDebugString(strViolationMessage, 4, 255, 255, 255);
 	outputServerLog(strViolationMessage);
 	fileWrite(uFileHandle, strViolationMessage.."\n");
-	fileFlush(uFileHandle);
 	fileClose(uFileHandle);
 end
 
@@ -39,7 +38,6 @@ function logAction(strMessage)
 	outputDebugString(strActionMessage, 4, 255, 255, 255);
 	outputServerLog(strActionMessage);
 	fileWrite(uFileHandle, strActionMessage.."\n");
-	fileFlush(uFileHandle);
 	fileClose(uFileHandle);
 end
 
--- a/[admin]/security/logging.lua
+++ b/[admin]/security/logging.lua
@@ -1,0 +1,65 @@
+-- log messages triggered by player
+function logViolation(uPlayer, strMessage)
+	local strPlayerName, strPlayerIP, strPlayerSerial = getPlayerName(uPlayer), getPlayerIP(uPlayer), getPlayerSerial(uPlayer);
+	local strLogFileName = "violations.txt";
+	local uFileHandle = fileExists(strLogFileName) and fileOpen(strLogFileName);
+	
+	if(not uFileHandle) then
+		uFileHandle = fileCreate(strLogFileName);
+		fileFlush(uFileHandle);
+	else
+		fileSetPos(uFileHandle, fileGetSize(uFileHandle));
+	end
+	
+	local strViolationMessage = getDateTime().." CLIENT: "..strPlayerName.." | IP: "..strPlayerIP.." | SERIAL: "..strPlayerSerial.." | "..strMessage;
+	
+	outputDebugString(strViolationMessage, 4, 255, 255, 255);
+	outputServerLog(strViolationMessage);
+	fileWrite(uFileHandle, strViolationMessage.."\n");
+	fileFlush(uFileHandle);
+	fileClose(uFileHandle);
+end
+
+
+
+-- log messages without player element
+function logAction(strMessage)
+	local strLogFileName = "actions.txt";
+	local uFileHandle = fileExists(strLogFileName) and fileOpen(strLogFileName);
+	
+	if(not uFileHandle) then
+		uFileHandle = fileCreate(strLogFileName);
+		fileFlush(uFileHandle);
+	else
+		fileSetPos(uFileHandle, fileGetSize(uFileHandle));
+	end
+	
+	local strActionMessage = getDateTime().." "..strMessage;
+	
+	outputDebugString(strActionMessage, 4, 255, 255, 255);
+	outputServerLog(strActionMessage);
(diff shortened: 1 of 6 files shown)


Fernando-A-Rocha:
You are basically implementing these principles 
https://wiki.multitheftauto.com/wiki/Script_security

Fernando-A-Rocha:
I suggested this on mta dev discord and people discussed it. Some said it is not enough to make a server secure, that is correct, it's a foundation as PLASMA stated here in this PR. I believe it makes sense to have this resource in the default resources pack. It's essential nowadays on all servers.

Fernando-A-Rocha:
Previous discussion here: https://discord.com/channels/801330706252038164/801407272483291136/1248315073844220084

ffsPLASMA:
Yes its not a solution to cheating, nor should it be considered. its just a base with some useful events to catch potential abusive players and build upon it. I always wondered why nothing of such sorts were included by default.

[admin]/security/logging.lua
```diff
@@ -0,0 +1,65 @@
...
+	else
+		fileSetPos(uFileHandle, fileGetSize(uFileHandle));
+	end
+	
+	local strViolationMessage = getDateTime().." CLIENT: "..strPlayerName.." | IP: "..strPlayerIP.." | SERIAL: "..strPlayerSerial.." | "..strMessage;
+	
+	outputDebugString(strViolationMessage, 4, 255, 255, 255);
+	outputServerLog(strViolationMessage);
+	fileWrite(uFileHandle, strViolationMessage.."\n");
+	fileFlush(uFileHandle);
```
Comment (Nico8340) [inline [admin]/security/logging.lua:19]: There is no need to use the `fileFlush` function because you close the file immediately

[admin]/security/logging.lua
```diff
@@ -0,0 +1,65 @@
...
+	else
+		fileSetPos(uFileHandle, fileGetSize(uFileHandle));
+	end
+	
+	local strActionMessage = getDateTime().." "..strMessage;
+	
+	outputDebugString(strActionMessage, 4, 255, 255, 255);
+	outputServerLog(strActionMessage);
+	fileWrite(uFileHandle, strActionMessage.."\n");
+	fileFlush(uFileHandle);
```
Comment (Nico8340) [inline [admin]/security/logging.lua:42]: There is no need to use the `fileFlush` function because you close the file immediately (same as above)

[admin]/security/logging.lua
```diff
@@ -0,0 +1,65 @@
...
+	local iMinute 		= tblRealTime.minute;
+	local iSecond 		= tblRealTime.second;
+	
+	if(iDay < 10) then iDay = "0"..iDay end;
+	if(iMonth < 10) then iMonth = "0"..iMonth end;
+	if(iHour < 10) then iHour = "0"..iHour end;
+	if(iMinute < 10) then iMinute = "0"..iMinute end;
+	if(iSecond < 10) then iSecond = "0"..iSecond end;
+	
+	return "["..tostring(iDay).."."..tostring(iMonth).."."..tostring(iYear).." - "..tostring(iHour)..":"..tostring(iMinute)..":"..tostring(iSecond).."]";
```
Comment (Nico8340) [inline [admin]/security/logging.lua:64]: How about using `string.format` instead?

[admin]/security/players.lua
```diff
@@ -0,0 +1,58 @@
...
+
+
+-- https://wiki.multitheftauto.com/wiki/OnPlayerNetworkStatus
+-- gets triggered when connection from server to a client is interrupted
+function clientNetworkStatus(iStatus, iTicks)
+	if(iStatus == 0) then
+		logViolation(source, "Network interruption has began after "..iTicks.." ticks");
+	elseif(iStatus == 1) then
+		logViolation(source, "Network interruption has stopped after "..iTicks.." ticks");
+	end
```
Comment (Nico8340) [inline [admin]/security/players.lua:56]: How about using a simple conditional check here?
```suggestion
logViolation(source, "Network interruption has " .. ((iStatus == 0) and "began" or "stopped") .. " after " .. iTicks .. " ticks")
```

Comment (ffsPLASMA) [reply [admin]/security/players.lua:56]: I used branch so people can further decide on what to do in case this event triggers.

[admin]/security/logging.lua
Comment (ffsPLASMA) [reply [admin]/security/logging.lua:64]: dunno, looks like cosmetic bias depending on different people. it works fine afterall.

Comment (Nico8340) [reply [admin]/security/logging.lua:64]: > dunno, looks like cosmetic bias depending on different people. it works fine afterall.

If you prefer it that way, you can keep it

[admin]/security/players.lua
Comment (Nico8340) [reply [admin]/security/players.lua:56]: > I used branch so people can further decide on what to do in case this event triggers.

It's fair enough

[admin]/security/logging.lua
Comment (ffsPLASMA) [reply [admin]/security/logging.lua]: removed

[admin]/security/logging.lua
Comment (ffsPLASMA) [reply [admin]/security/logging.lua]: removed

jlillis (changes requested):
I get the intent here but I really don't see this as a useful resource. If all it is doing is logging potentially problematic behavior. For this to be useful, it should really be integrated into admin/admin2 and take some action by default.

Dutchman101:
> I get the intent here but I really don't see this as a useful resource. If all it is doing is logging potentially problematic behavior. For this to be useful, it should really be integrated into admin/admin2 and take some action by default.

This is just the beginning, after we see results from varying servers and gamemodes, the logging sensitivity may be tweaked to the point it's suitable for violation to lead to an action (or specific violation subtypes, definable by the server owner, as appropiate for their gamemode). Also, if we add universal, generic security checks that aren't mainly to raise an alert but to make a certain cheat effect impossible or violate on it, the code will grow expotentially as well as it will do down the line anyways, as again this is a very early version and there's much more to do here; it's better in a dedicated resource than bloating admin, or having to mind its structure when expanding it.

Really more of a "good to have, why not" thing at this point, and it can become really of value as it gets expanded to cover way more types of things.

PR #534, opened 2024-08-11
editor_main: Improved undo/redo functionality

q8X:


Fernando-A-Rocha:
What exactly does this improve?

q8X:
- Fixes the action/index table doesn't get reset after open/new map
- Fixes the insertAction function, removes the wrong index when you undo an object. create > delete > undo > delete > undo 
- Removed unnecessary variables and checks that break indexing

https://github.com/user-attachments/assets/92785c0d-43d1-40cd-b930-95d28e986788

PR #539, opened 2024-08-27
editor: Fix UI appearing during map testing when "collision patches" enabled

### LINKED ISSUE #532: unexpected editor UI appearing with "collision patches" enabled during map testing
edtior_gui

**Describe the bug**
when the collision patches feature is enabled during map testing, the editor UI show up, this behaviour isn't supposed to be happened

**To Reproduce**
1. Enable the "collision patches" feature in the editor settings
2. Begin a full test
3. See the editor UI appearing during the test

**Expected behavior**
the editor UI shouldn't be visible during map testing with the "collision patches" feature enabled

**Screenshots**
mta-screen_2024-07-27_19-23-23

**Version**
- Multi Theft Auto v1.6-release-22650
- Latest MTA default resources

q8X:
Fixes #532, Currently the map editor shows the UI when collision patches are enabled during map testing, this caused by https://github.com/multitheftauto/mtasa-resources/blob/80d98fdc7daf7748756a7726b3261362ad8bdc9c/%5Beditor%5D/editor_gui/client/options_gui.lua#L127-L136

### DIFF
--- a/[editor]/editor_gui/client/test.lua
+++ b/[editor]/editor_gui/client/test.lua
@@ -142,7 +142,7 @@ function disableColPatchInTesting()
 	
 	-- Disable
 	guiCheckBoxSetSelected(dialog.enableColPatch.GUI.checkbox, false)
-	confirmSettings()
+	doActions()
 end
 
 function enableColPatchAfterTesting()


jlillis:
Tested and working, thanks!

PR #540, opened 2024-08-28
[editor] Fix editor_gui test panel

chris1384:
Fixes: After testing a certain map with a gamemode (ex. race), you are unable to test the same map with the `<None>` option anymore.
Retains last testing gamemode upon reconnecting.


jlillis:
Tested and working, thanks!

PR #542, opened 2024-08-29
[editor] Toggleable rotation mechanic and improve threshold

### LINKED ISSUE #537: Editor: Object getting flipped when rotating
Editor

**Describe the bug**
setting the object rotation to 270 in the Y axis and 90 in Z axis, and rotate the X axis using (either local or world) modifier, cause the object to flip to the opposite

**To Reproduce**
1. Open map editor
2. Create an Object
3. Set object rotation to 270 in the Y axis and 90 in Z axis
4. Rotate the object using either local modifier (hold RCTRL and press arrow left or arrow right) or world modifier (hold LCTRL and press page up or page down)
5. Notice that the object flipped after the rotation

**Expected behavior**
The object should rotate smoothly without any flipping

**Screenshots**

https://github.com/user-attachments/assets/e971fe9e-4592-4a17-a160-9e04a6953718

**Version**
Multi Theft Auto v1.6-release-22684

chris1384:
Make the "`Enable rotation patches`" work as a toggle between the current rotation behaviour and the old 1.5.9 rotation. 
`editor_gui` field might need a renaming.
Also fixes #537

### DIFF
--- a/[editor]/editor_main/client/rotation.lua
+++ b/[editor]/editor_main/client/rotation.lua
@@ -61,6 +61,9 @@ function applyIncrementalRotation(element, axis, angle, world_space)
 	end
 	offset_quat[4] = math.cos(arad / 2)
 	
+	-- Get rotation patch userdata
+	local enableRotPatch = exports["editor_gui"]:sx_getOptionData("enableRotPatch")
+	
 	-- Get current rotation
 	local cur_quat
 	if elementQuat[element] then
@@ -71,7 +74,7 @@ function applyIncrementalRotation(element, axis, angle, world_space)
 		if euler_rot[1] == 0 and euler_rot[2] == 0 and euler_rot[3] == 0 then
 			-- Is there a fix and are rotation patches enabled
 			local id = getElementModel(element)
-			if rotationFixes[id] and exports["editor_gui"]:sx_getOptionData("enableRotPatch") then
+			if rotationFixes[id] and enableRotPatch then
 				-- Rotate from the fix
 				cur_quat = {unpack(rotationFixes[id])}
 			else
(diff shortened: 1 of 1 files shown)


dnl-k:
That's quite redundant if you ask me

chris1384:
> That's quite redundant if you ask me

I don't think anyone uses the current LCTRL for X and Y transforms, players being used to the old mechanic can safely use the newer editor without experiencing these bind changes, instead of being stuck to the older, unstable ones.

[editor]/editor_main/client/rotation.lua
```diff
@@ -130,10 +140,10 @@ function getEulerFromQuat(quat)
     local q0, q1, q2, q3 = quat[1], quat[2], quat[3], quat[4]
     local treshold = q0 * q2 - q3 * q1
 	
-    if treshold > 0.499 then
+    if treshold > 0.499999999 then
         return {math.deg(2 * math.atan2(q1, q0)), 90, 0}
-    elseif treshold < -0.499 then
-        return {math.deg(-2 * math.atan2(q1, q0)), -90, 0}
+    elseif treshold < -0.499999999 then
```
Comment (jlillis) [inline [editor]/editor_main/client/rotation.lua:145]: Maybe define this "magic number" in a variable at the top of the file?

jlillis (approved):
Only one minor nitpick. I have no issue with changing the default rotation behavior but others may disagree, so I'd rather wait for more feedback (a week or two max).

dnl-k:
imo the default rotation behaviour should be the one before the "Rotation improvements"
It should've never been changed.

The "improved" one can be toggleable via said setting

[editor]/editor_main/client/rotation.lua
Comment (chris1384) [reply [editor]/editor_main/client/rotation.lua]: all done

jlillis:
Haven't had anyone disagree with the change in behavior so I think we'll merge this. If anyone complains in the future we may need to re-evaluate.

PR #543, opened 2024-08-29
[editor] Fix duplicate map meta entries resulting in editor suspending.

### LINKED ISSUE #538: Editor: Duplicate the same filename in meta.xml causes editor to stop
Editor

**Describe the bug**
Duplicating the same filename in meta.xml causes the editor to stop, It's hard to find this bug when you handle a lot of maps, and rarely editor duplicate it in editor_test/dump for some reason

**To Reproduce**
1. Open meta.xml in any map
2. Duplicate any filename in the same path and save the file
3. Open the map in editor
4. Start a full test
5. See error

**Expected behavior**
editor should ignore any duplicate filenames in the same path

**Screenshots**
image
image

**Version**
Multi Theft Auto v1.6-release-22684

chris1384:
Fixes #538

Screenshot_24

This does not check for multiple files with different attributes (ex. scripts type that use both `client` and `server` instead of `shared`).

### DIFF
--- a/[editor]/editor_main/server/resourcehooks.lua
+++ b/[editor]/editor_main/server/resourcehooks.lua
@@ -101,7 +101,9 @@ function getResourceFiles ( resource, fileType )
 		local file = xmlNodeGetAttribute ( node, "src" )
 		local otherAttributes = xmlNodeGetAttributes ( node )
 		otherAttributes.src = nil
-		if not fileAttributes[file] then
+		if fileAttributes[file] then
+			outputDebugString("getResourceFiles: Found duplicate meta entry in '".. resource .."' (".. fileType .. " - ".. file .. ")")
+		else
 			fileAttributes[file] = otherAttributes
 			table.insert ( files, file )
 		end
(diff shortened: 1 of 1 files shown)


jlillis:
I'm not sure ignoring duplicate entries is the best idea - there should a warning shown to the user informing them of the issue. If `editor` is duplicating files that means there is a bug that needs to be fixed.

chris1384:
dup

The editor does exactly what the map meta contains, but it fails at `fileCopy` if there's a file already copied.
I wouldn't say that the editor is at fault here, it's either the users fault or an another resource with `general.ModifyOtherObjects` permissions and writes extra lines in the meta. I can't think of a better approach for this so feel free to suggest anything. :d

jlillis:
> I wouldn't say that the editor is at fault here, it's either the users fault or an another resource with `general.ModifyOtherObjects` permissions and writes extra lines in the meta.

I think you're probably right. If this breaks anything at least we'll have a debug warning to trace it with.

PR #546, opened 2024-09-03
editor: Fix EDF custom elements rotation issues on undo/manual Adjustment

q8X:
This PR fixes EDF rotation issue when setting the rotation manually or undoing an element, this could happen if you:
1. rotate > undo  > rotate after undoing
2. rotate > set rotation (F3) > rotate

This bug caused by a change in commit #400 

### **A video explains how the bug works**
I used the race spawnpoint as an example

https://github.com/user-attachments/assets/8207c082-0cd2-4ed1-957a-93c9ef0904b6

### DIFF
--- a/[editor]/move_keyboard/move_keyboard.lua
+++ b/[editor]/move_keyboard/move_keyboard.lua
@@ -190,9 +190,9 @@ local function onClientRender_keyboard()
 				else
 					local tempRotX, tempRotY, tempRotZ = rotX, rotY, rotZ
 					if (not tempRotX) then return false end
-
+					
 					local yaw, pitch, roll = 0, 0, 0
-
+					
 					-- yaw
 					if (getCommandTogSTATE("element_move_right")) then
 						yaw = speed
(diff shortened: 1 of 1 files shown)


[editor]/move_keyboard/move_keyboard.lua
```diff
@@ -190,9 +190,9 @@ local function onClientRender_keyboard()
 				else
 					local tempRotX, tempRotY, tempRotZ = rotX, rotY, rotZ
 					if (not tempRotX) then return false end
-
```
Comment (jlillis) [inline [editor]/move_keyboard/move_keyboard.lua:193]: Looks like you have some unintended whitespace changes. Mind cleaning these up?

jlillis (changes requested):
Everything looks good, just a minor nitpick about some whitespace changes.

[editor]/move_keyboard/move_keyboard.lua
Comment (q8X) [reply [editor]/move_keyboard/move_keyboard.lua]: done

jlillis (approved):

PR #552, opened 2024-09-26
internetradio: Fix playSound3D and track name showing in other dimensions

mateo-14:
Fix bugs caused by playing the sound in dimensions other than 0
- The sound created by playSound3D didn't move to the player's dimension, so the player could not hear it
- Track name wasn't visible in other dimensions


Dutchman101:
Thanks @mateo-14

PR #553, opened 2024-09-28
[editor] Fix debugscript warnings when testing an unopened map

chris1384:
Reproducing the bug:
`1. Start editor`
`2. Test`

Second check is added to avoid the other errors caused by `copyResourceFiles()`.
Set as draft after finding out `currentMapSettings` causes `edf` script errors.


chris1384:
Closing this as it will be included in an upcoming PR.

PR #554, opened 2024-10-08
Race edf: Fix automatic nextid assignment breaking

lotsofs:
Duplicating a checkpoint with a nextid set would point that clone to the same checkpoint which resulted in multiple checkpoints pointing to another resulting in stray checkpoints.

When deleted checkpoints are in play and a new checkpoint was created, it would be automatically pointed to by said deleted checkpoint instead of an existing one, resulting in stray checkpoint sequences.

Fixed both. New checkpoints are now placed after the existing sequence as expected.


Nico8340 (commented):
LGTM

PR #556, opened 2024-10-11
editor: multiple fixes & improvements related to LODs

### LINKED ISSUE #555: Map editor LOD_MAP table is missing some LOD models
editor_main => server => mapEditorScriptingExtension_s.lua

https://github.com/multitheftauto/mtasa-resources/blob/d4c04db009cdd68913fdb47bbc73acd91e63f981/%5Beditor%5D/editor_main/server/mapEditorScriptingExtension_s.lua#L78

**Describe the bug**
This table is missing certain `object model ID => LOD model ID` associations.

e.g. ID `3589` ([`compfukhouse3`](https://dev.prineside.com/en/gtasa_samp_model_id/model/3589-compfukhouse3/))'s LOD model is `3592` (`LODpfukhouse3`)

**To Reproduce**
The correct LOD model of a world object / building is actually provided in `processLineOfSight` ("_This function casts a ray between two points in the world, and tells you information about the point that was hit, if any._"), which is used by the Map Editor in editor_main => client => main.lua @ `processCursorLineOfSight()` obtaining `buildingId` and `buildingLOD`.

**Expected behavior**
LOD_MAP should include all Object=>LOD model IDs in GTA:SA.

**Screenshots**

**Version**
MTA LOD Table [object] = [lodmodel]

Version of Apr 11, 2023 (last change as of submitting this issue)

**Additional context**
I created a Python script to parse all of the game's IPL files (including the binary IPLs converted to text), and create a dictionary of object model ID => LOD object model ID according to the info in https://gtamods.com/wiki/LOD#GTA_San_Andreas_.26_GTA_IV . I obtained the same LOD_MAP table in this experiment.

I noticed that some LOD Models like `LODpfukhouse3 (3592)` aren't actually linked by index by any object model in the same IPL, which is why the script did not find them... Hmm...

Fernando-A-Rocha:
**Status**: waiting for new MTA feature https://github.com/multitheftauto/mtasa-blue/pull/3831

- [X] editor_main: Makes useLODs setting enabled by default. It's a feature that the game uses for world objects, there is no reason to not have it enabled for map objects in new maps.
- [X] editor_main: Modernizes the map scripting extensions script: gets rid of the clientside **engineSetModelLODDistance to 300**, which did not make sense. Let's keep the default model LOD distances.
- [X] editor_main: Makes use of the new `getObjectLowLODModel` MTA function to obtain the LLOD object ID of any HLOD object ID. The previous LOD_TABLE was even incomplete! Rockstar hid some object=>lod associations in binary IPL files, which have been identified and fixed in [MTA PR 3831](https://github.com/multitheftauto/mtasa-blue/pull/3831). Fixes #555
- [X] edf & editor_main: Fixes LOD models not creating for objects during mapping


Fernando-A-Rocha:
Tested!
Ready for review, im satisfied with the result

chris1384:
I don't think that useLODs should be a good idea to be true as default, since most players are not aware of what LODs do in the first place.
And this may affect players with low-end PCs.
Other than that, seems like a good refactor overall.

Fernando-A-Rocha:
The LOD object model only loads when the object becomes far away, replacing the normal object (or so it should). I doubt it's bad in terms of performance...

[editor]/editor_main/server/dumpxml.lua
```diff
@@ -239,3 +234,73 @@ function getMapElementData ( element )
...
+			end
+		end
+	end
+end
+addEventHandler("onResourceStart", resourceRoot, onResourceStartOrStop, false)
+addEventHandler("onResourceStop", resourceRoot, onResourceStartOrStop, false)
+
+-- MTA LOD Table [object] = [lodmodel]]
+
+SCRIPTING_EXTENSION_CODE = SCRIPTING_EXTENSION_CODE .. "\n" .. getLodsLuaTableString()
```
Comment (Dark-Dragon) [inline [editor]/editor_main/server/dumpxml.lua:306]: Given this file would end up in hundreds or thousands of map resources in the future, I think it would be reasonable to attempt to cut down on the amount of information in it. It wouldn't be unfeasible to only add the entries relevant to the map in question, would it? IE only add the information about LODs for objects that are actually used in the map at the time it's saved. If the author adds more objects in a later revision those would need to be checked and added then.

[editor]/editor_main/server/dumpxml.lua
```diff
@@ -239,3 +234,73 @@ function getMapElementData ( element )
...
+					usedLODModels[lodModel] = true
+				end
+			end
+		end
+	end
+end
+addEventHandler("onResourceStart", resourceRoot, onResourceStartOrStop, false)
+addEventHandler("onResourceStop", resourceRoot, onResourceStartOrStop, false)
+
+-- MTA LOD Table [object] = [lodmodel]]
```
Comment (Dark-Dragon) [inline [editor]/editor_main/server/dumpxml.lua:304]: Needs one more closing ] at the end, two to close the multi-line string and one to encapsulate 'lodmodel' in the final file

`-- MTA LOD Table [object] = [lodmodel]]]`

[editor]/editor_main/server/dumpxml.lua
Comment (Fernando-A-Rocha) [reply [editor]/editor_main/server/dumpxml.lua:306]: Including only the lod models of objects used by the map doesn't optimize much. It creates the problem that you described, if the map is altered changing it's objects, this script would have to be updated.

Comment (Fernando-A-Rocha) [reply [editor]/editor_main/server/dumpxml.lua:306]: So imo it's better to leave it as I did

Comment (Dark-Dragon) [reply [editor]/editor_main/server/dumpxml.lua:306]: Ok this would be something for a future pull request, but it seems we are currently adding this file to every map even if it doesn't use any of the features that require it. This seems like waste of storage space that we should be able to avoid.

To be honest it should probably be on a separate resource (maybe something like mapmanager) to provide the required basic functionality on map launch and if not, I'd say the least we can do is only add the file if it's required and only add the required components. Even though the output file is only 70kb that's 70kb extra for every single map, many of which would probably be smaller than 10kb otherwise.

We can consider this resolved for the sake of this PR but it is something worth discussing alternative options for. I'll be opening an issue in a bit.

Comment (Fernando-A-Rocha) [reply [editor]/editor_main/server/dumpxml.lua:306]: For the sake of this PR let's not complicate. But I agree with you: the best solution would be to have 1 maploader resource to handle all maps. Makes no sense to have "map scripting extension" scripts in EVERY map resource.

Fernando-A-Rocha:
Actually, keep this on hold.
I'm gonna make a mtasa-blue function to return the new LOD_TABLE :-) It will be useful to have this in MTA itself

Fernando-A-Rocha:
Updated to use new feature from https://github.com/multitheftauto/mtasa-blue/pull/3831

Fernando-A-Rocha:
Work on this PR is paused until we figure out https://github.com/multitheftauto/mtasa-blue/pull/3831

q8X:
i cant test this PR due to these errors

```
[01:29:56] ERROR: [editor]\edf\edf.lua:33: attempt to call global 'getObjectLowLODModel' (a nil value)

[01:29:56] ERROR: [editor]\editor_main\server\interface.lua:11: call: failed to call 'edf:edfRepresentElement'

[01:29:56] WARNING: [editor]\edf\edf.lua:825: Bad argument @ 'getElementData' [Expected element at argument 1, got boolean]
[01:29:56] WARNING: [editor]\editor_main\server\IDhandler.lua:45: Bad argument @ 'getElementType' [Expected element at argument 1, got boolean]

[01:29:56] WARNING: [editor]\editor_main\server\IDhandler.lua:47: Bad argument @ 'getElementID' [Expected element at argument 1, got boolean]

[01:29:56] ERROR: [editor]\editor_main\server\util.lua:119: [editor]\editor_main\server\IDhandler.lua:71: attempt to concatenate local 'idString' (a boolean value) [string "?"]

[01:29:56] ERROR: [editor]\editor_main\server\util.lua:121: [editor]\editor_main\server\IDhandler.lua:71: attempt to concatenate local 'idString' (a boolean value)
```

Fernando-A-Rocha:
I will remake this PR. I don't think there will be an addition of a function to get the LOD of a model in mtasa-blue. So we need to do this in Lua.

I am thinking of a new resource. Map utilities or something like that. To handle LOD etc.

PR #557, opened 2024-10-11
editor: change useLODs map setting to true by default

Fernando-A-Rocha:
There is no reason to disable object LODs by default for maps. It's an amazing feature that makes maps look better from a far distance, and it's used in the original game for all world objects... Therefore it should be enabled by default.

See the discussion here for more context https://github.com/multitheftauto/mtasa-resources/pull/556#issuecomment-2407198020


Fernando-A-Rocha:
merge after #556 :-)

then you can spawn a random house object, and load the map normally then fly away to see that the LOD loads at an awesome distance.

Fernando-A-Rocha:
Closing this PR because i wanna do more edits in #556, i'll incorporate this in that PR.

PR #566, opened 2024-10-19
Fix chat messages sent twice

### LINKED ISSUE #460: Chat messages sent twice

Fernando-A-Rocha:
Fixes #460 

playercolors will no longer handle chat, as this is handled in the gamemode resource itself (freeroam), playercolors is just an extra resource.

@Dark-Dragon @Nico8340


Fernando-A-Rocha:
Other game modes that handle and output chat to chatbox are:
- TDM
- CTV
- Assault

Then, `admin` and `webstats` also handle onPlayerChat but don't alter its output.

Dark-Dragon:
Appropriate fix for while we wait for the large scale changes discussed in #463

Fernando-A-Rocha:
Exactly, this can be merged now.

[gameplay]/playercolors/playercolors.lua
```diff
@@ -20,18 +20,5 @@ addEventHandler("onResourceStart", resourceRoot, setAllPlayerColors)
 addEventHandler("onGamemodeMapStart", root, setAllPlayerColors) -- mapmanager resets player colors to white when the map ends
 addEventHandler("onResourceStop", resourceRoot, setAllPlayerColors)
 
-addEventHandler('onPlayerChat', root,
-	function(msg, type)
-		if type == 0 then
```
Comment (jlillis) [inline [gameplay]/playercolors/playercolors.lua:25]: I think it would be better to just add an early return if `freeroam` is running, as this would break functionality on resources other than `freeroam`. Not that it's important functionality, I guess.

jlillis (changes requested):

[gameplay]/playercolors/playercolors.lua
Comment (Fernando-A-Rocha) [reply [gameplay]/playercolors/playercolors.lua:25]: I can't think of a reason you would need playercolors to handle chat for you, running alongside other resources... 🤔 To me, it's a bug that this PR fixes.

Fernando-A-Rocha:
@jlillis do you agree with my comment above?

[gameplay]/playercolors/playercolors.lua
Comment (jlillis) [reply [gameplay]/playercolors/playercolors.lua:25]: You would need it to if you expect the player's nametag colors to appear in chat in gamemodes other than `freeroam`.

Comment (Fernando-A-Rocha) [reply [gameplay]/playercolors/playercolors.lua:25]: > I think it would be better to just add an early return if `freeroam` is running, as this would break functionality on resources other than `freeroam`. Not that it's important functionality, I guess.

Ok

Fernando-A-Rocha:
There you have it @jlillis

jlillis (approved):

PR #570, opened 2024-10-29
added state to unbind toggleMap from being bound to F2

### LINKED ISSUE #535: freeroam: Map key bind interferes with race editor help
freeroam, editor / race

**Describe the bug**
https://github.com/multitheftauto/mtasa-resources/commit/4083f3adcf44c0b61f2ebbc53edfeac24005fdaf Added F2 to show the map in the freeroam resource. The race editor help thingy in the race gamemode uses F2 as show/hide. Since the editor relies on freeroam and both need to be present at the same time, a different key for either the map or the help thingy should be considered.

**To Reproduce**
1. Start your local map editor
2. Start a map with the race gamemode
3. Try hiding the help thingy by pressing F2
4. It'll show the map.

**Expected behavior**
One key shouldn't be bound to two functions.

**Version**
1.6

MittellBuurman:
Fixes #535

- Added a check to see if the race recource is running when freeroam is running. Disabling/unbinding the toggleMap feature from the client (since you're racing, and likely testing the map, having a map spawn selector doesn't seem useful here).

### DIFF
--- a/[gameplay]/freeroam/fr_client.lua
+++ b/[gameplay]/freeroam/fr_client.lua
@@ -2243,8 +2243,18 @@ addEventHandler('onClientResourceStart', resourceRoot,
 		createWindow(wndMain)
 		hideAllWindows()
 		bindKey('f1', 'down', toggleFRWindow)
+		bindKey('f2', 'down', toggleMap)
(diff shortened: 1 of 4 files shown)


MittellBuurman:
Closed PR due to author commit conflicts

PR #571, opened 2024-11-01
freeroam: bugfix on fr_client.lua onExitVehicle function

kutlu35:
There is no point of putting 'elseif' on that spesific place, it causes problems with anti-ramming system since it actually blocks the setVehicleGhost(vehicle,hasDriverGhost(vehicle)), it should be 'end if' instead. That way when driver of the vehicle exits the vehicle, ghostmode vehicle can be back to it's normal state.

VIDEO of the bug: https://youtu.be/ZRFNxSeI1Eo

### DIFF
--- a/[gameplay]/freeroam/fr_client.lua
+++ b/[gameplay]/freeroam/fr_client.lua
@@ -2123,7 +2123,9 @@ function onExitVehicle(vehicle,seat)
 		hideControls(wndMain, 'repair', 'flip', 'upgrades', 'color', 'paintjob', 'lightson', 'lightsoff')
 		closeWindow(wndUpgrades)
 		closeWindow(wndColor)
-	elseif vehicle and seat == 0 then
+	end
+		
+	if vehicle and seat == 0 then
 		if source and g_PlayerData[source] then
 			setVehicleGhost(vehicle,hasDriverGhost(vehicle))
 		end


Dark-Dragon:
While that bit in the code seems to be structured more confusingly than necessary, I believe the point is to have the bottom part only apply for cases where it's not the vehicle of the local player and only for the "onClientPlayerVehicleExit" event. Whether that's important for the functionality of the script I'd need to test, but I wouldn't really agree that there is 'no point' unless you can provide more of an explanation.

kutlu35:
> While that bit in the code seems to be structured more confusingly than necessary, I believe the point is to have the bottom part only apply for cases where it's not the vehicle of the local player and only for the "onClientPlayerVehicleExit" event. Whether that's important for the functionality of the script I'd need to test, but I wouldn't really agree that there is 'no point' unless you can provide more of an explanation.

First part of the function is for destroying buttons such as 'repair', 'flip', 'color' when any localPlayer exits vehicle, bottom part of the function is actually for changing the data of the vehicle when the DRIVER of the vehicle (seat == 0) exits the vehicle. For example we got "setVehicleGhost(vehicle,hasDriverGhost(vehicle))" there, this makes ghost vehicle normal vehicle again when a player that ticked 'anti ram' button exits the vehicle. 

I encountered this bug when i was making a "anti-glue(make other players can't glue to my vehicle)" system for freeroam. The system was working perfectly except when i got out the vehicle, vehicle kept it's data unglueable(vehicle should be glueable by other players again when you exit the vehicle).

Also you can check onEnterVehicle function, it has same structure and it got "end if" instead of "elseif" and it works perfectly fine with my antiglue system (For example if a vehicle is glueable and has no driver  and then if a player that turned on anti-glue enters the vehicle, vehicle gets unglueable automatically).

kutlu35:
> While that bit in the code seems to be structured more confusingly than necessary, I believe the point is to have the bottom part only apply for cases where it's not the vehicle of the local player and only for the "onClientPlayerVehicleExit" event. Whether that's important for the functionality of the script I'd need to test, but I wouldn't really agree that there is 'no point' unless you can provide more of an explanation.

First part of the function is for destroying freeroam panel buttons such as 'repair', 'flip', 'color' when localPlayer exits vehicle, bottom part of the function is actually for changing the data of the vehicle when the DRIVER of the vehicle (seat == 0) exits the vehicle. For example we got "setVehicleGhost(vehicle,hasDriverGhost(vehicle))" there, this makes ghost vehicle normal vehicle again when a player that ticked 'anti ram' button exits the vehicle, but it won't actually happen. So this bug actually messes up the  anti-ramming system. Also you can check the onEnterVehicle function which has same structure but it has "end if" instead of "elseif" and it works perfectly fine.

Here is a video:
https://youtu.be/ZRFNxSeI1Eo

Dark-Dragon:
With the added context your change seems to restore the intended behavior.

jlillis (approved):

PR #576, opened 2024-11-17
editor: adding missing things

THEGizmoOfficial:
Fixes for the following things for object **ID: 5993**
- Added missing object ID to the map editor list
- Added missing collision to editor colpatch


Fernando-A-Rocha:
Are there any more missing models from that objects list? Comparing with GTA SA IDE files

THEGizmoOfficial:
> Are there any more missing models from that objects list? Comparing with GTA SA IDE files

I'm not sure but perhaps

PR #579, opened 2024-11-21
Update fr_server.lua

shadylua:
I noticed a bug in the freeroam game mode,  

        <setting name=“*chat/blockRepeatMessages” value=“true” />

but I was still able to spam the chat, I looked at the line of code and saw that something was wrong and fixed it


shadylua:
fixed lint **[gameplay]/freeroam/fr_server.lua:3:7: variable chatTime is mutated but never accessed**

PR #585, opened 2024-12-14
superman: further improvements

ds1-e:
- Removed extra table overhead from functions
-> Localized few of them
- Moved configurable vars to separate .lua
- Moved vectors to separate .lua
-> Use simple tables from now on
-> Probably fixes rare error below (for some reason not breaking anything, just annoying script log)
-> ```attempt to call method 'Normalize' (a nil value)```
* Different, and more effective approach for getting flying players around; instead of searching for players in radius of 300, script now gets only those who are streamed in.
-> This also ensures that players are in same interior & dimension.


Dutchman101:
Thanks a lot, looks good

PR #588, opened 2024-12-18
Add recently introduced events, name and explosion/projectile checking

ffsPLASMA:
invalid playername check on connect
added onPlayerChangesWorldSpecialProperty
added onPlayerChangesProtectedData
added interval checks for onPlayerProjectileCreation, onExplosion, onVehicleExplode with timer interval and thresholds


Nico8340 (commented):
Code overall looks good, but you should add an exception list to special property change, to only flag specific ones

ffsPLASMA:
> Code overall looks good, but you should add an exception list to special property change, to only flag specific ones

I rather keep it up to the individual to add exceptions for special properties.

Nico8340:
> I rather keep it up to the individual to add exceptions for special properties.

I understand the intention, but this event works in such a way that it is called every time a special property changes, even when connecting to the server, and it will insert a lot of unnecessary logs.

ffsPLASMA:
> > I rather keep it up to the individual to add exceptions for special properties.
> 
> I understand the intention, but this event works in such a way that it is called every time a special property changes, even when connecting to the server, and it will insert a lot of unnecessary logs.

I mean this could be said about any event added in this resource. I dont see a point to add a property whitelist for whatever world property I deem worthy or not.

PR #590, opened 2025-01-04
mapfixes: add model transparency fixes

Fernando-A-Rocha:
Fix for https://github.com/multitheftauto/mtasa-blue/issues/2074

Idea by @FileEX 

WIP


Fernando-A-Rocha:
Sets `draw_last` model flag to model ID 3089 so its glass transparency works correctly and you can see through it.

BEFORE:

image

AFTER:

image

ghost (commented):
Good job!

Fernando-A-Rocha:
This is a draft because I used engineRestreamWorld() to reload the model but this is overkill, and it also affects the player by making it get stuck and move on the ground when spawning if done at the same time as map fixes loads (happens when you join freeroam).

Me and @FileEX are gonna find a solution

Fernando-A-Rocha:
because of the model flags

Fernando-A-Rocha:
Maybe this can be used to re-stream the model after changing its model flags
https://github.com/multitheftauto/mtasa-blue/pull/4357 instead of re-stream world

Fernando-A-Rocha:
#4357 in mtasa-blue was merged! So we can soon advance with these model fixes as we now have model restream functionality!

FileEX:
What's the progress?

ArranTuna:
@Fernando-A-Rocha I've made the necessary changes to use engineRestreamModel does it look good to you?

Now when I restart it, instead of the whole world flashing, just the door flashes.

Fernando-A-Rocha:
LGTM GJ !

Fernando-A-Rocha:
Awaiting 1.7

ArranTuna:
I didn't know that new commits to this repo was being sent to resources of 1.6 installers. There doesn't seem to be any way to re-open a PR? Also could add a version that works with 1.6 and 1.7 quite easily:

if engineRestreamModel then
 engineRestreamModel(model)
else
  engineRestreamWorld()
end

Fernando-A-Rocha:
ye youre right we can make a version that is compatible with 1.6 and 1.7 at the same time !

```lua
local isEngineRestreamModelAvailable = type(engineRestreamModel) == "function"
```

PR #591, opened 2025-01-04
Refactor headshot

Nico8340:
Refactors the headshot resource, breaks it into several parts, provides runtime modifiable settings, detailed documentation and examples.


[gameplay]/headshot/meta.xml
```diff
@@ -1,11 +1,12 @@
...
-			group="General"
-			accept="false,true"
-			desc="Should player lose his head when he gets headshot?" />
-	</settings>
-</meta>
+    <info name="Headshot" description="Headshot feature for enhanced combat realism and damage effect" author="Multi Theft Developers" version="1.0.0" type="script" />
+    <min_mta_version server="1.6.0" />
+
+    <settings>
+        <setting name="*decap" desc="determines whether the player becomes headless after a headshot" value="true" accept="false, true" />
```
Comment (Fernando-A-Rocha) [inline [gameplay]/headshot/meta.xml:6]: Value should be `"[true]"` and not `"true"` so that the script automatically converts to boolean, and you can avoid having the boolean convert/check methods in headshotSettings.lua

Fernando-A-Rocha (commented):
Overall good job!! But

[gameplay]/headshot/meta.xml
Comment (Fernando-A-Rocha) [reply [gameplay]/headshot/meta.xml:6]: See mapfixes example:
- https://github.com/multitheftauto/mtasa-resources/blob/master/%5Bgameplay%5D/mapfixes/scripts/server/main.lua
- https://github.com/multitheftauto/mtasa-resources/blob/master/%5Bgameplay%5D/mapfixes/meta.xml#L10

Comment (Fernando-A-Rocha) [reply [gameplay]/headshot/meta.xml:6]: The settings system is weird, but I think that way is correct ^^

Comment (Nico8340) [reply [gameplay]/headshot/meta.xml]: Changed

Lamaaar (commented):
Great job friendo!

Fernando-A-Rocha (commented):
👍🏼

Dutchman101:
Please remove all comments, headers (of where it's part of, etc), and documentation/scripting guide lines, and make it clean code only. Thanks for the effort put into these things but im only requesting this because i's not an established practise within other official resources, and to keep things as uniform as possible.

Also, in new resources, we now use this for meta.xml:
`<info author="MTA contributors (github.com/multitheftauto/mtasa-resources)" version="1.0"`

Nico8340:
Done @Dutchman101

PR #592, opened 2025-01-06
Update vehicles.xml

SpeedyFolf:
Added missing vehicles to map editor.

505 (Rancher Lure)
594 (RC Cam)
604 (Glendale Damaged)
605 (Sadler Damaged)

> [!NOTE]  
> Until https://github.com/multitheftauto/mtasa-blue/pull/3797 is merged, 543 (Sadler) and 605 (Sadler Damaged) will share the same name, but you can just click them to preview which is which.


Nico8340 (commented):
Very good

Kinimel:
I agree, your proposal seems very good, and it should also fix vehicles missing from the vehicle panel.

PR #593, opened 2025-01-06
Update skins.xml

SpeedyFolf:
Standardized skins.xml files for admin, admin2, editor_gui, and freeroam.

Added missing skins to admin2 and editor_gui, removed duplicate 103 (Baller Jacket) entries, and removed space and trailing comma from keywords.


Kinimel:
I agree you have a good point that it could be a problem of missing skins point of view and they should be reintegrated into the skins panel and not be missing from the skins panel

Nico8340 (commented):
Good

PR #596, opened 2025-01-22
Updated Flag of Syria

Zallb:
It has been already updated on many platforms (e.g. Steam https://steamcommunity.com/id/526556, also Android 15)

steamuserimages-a akamaihd

MTA is all about contributions from the community, it would make no sense to stick with the old flag (and wait 'till its update on the World Factbook first, for example)


Nico8340 (commented):
Looks good

PR #599, opened 2025-01-30
freeroam: fix false positive issue in `onPlayerTeleport` event and desync.

imfelipedev:
This PR fixes a false positive in the `onPlayerTeleport` event caused by the use of `setElementPosition` on the client side, as well as resolving desync issues resulting from improper use of the function on the wrong execution side (issue: https://github.com/multitheftauto/mtasa-blue/issues/3993).


Nico8340 (commented):
Looks good

PR #598, opened 2025-01-30
update getPedControlState calls to include localPlayer

Xenius97:
Fix warnings after updated getPedControlState (PR: https://github.com/multitheftauto/mtasa-blue/pull/3964)

PR #600, opened 2025-02-07
Fix freecam controlState

sousateew:
Someone forget to put `localPlayer` in getPedControlState... (it was causing bugs)


Nico8340:
Same as #598
Unnecessary, the bug has been fixed here: https://github.com/multitheftauto/mtasa-blue/pull/4008

PR #602, opened 2025-02-17
editor_main: adding missing objects to list

### LINKED ISSUE #601: When adding weapon models as objects to a map, it doesn't save their name

Something in [editor]

**Describe the bug**

When adding weapon models as objects to a map, it doesn't save their name in the .map file, neither their custom name like `Regular_Dildo` or model name like `gun_dildo1` (321), they will just be named like `object (1)` and `object (2)` and so on instead of `object (gun_dildo1) (1)` and `object (gun_dildo2) (1)` and so on.

**To Reproduce**

Delete the server folder from your MTA installation to make sure you have a fresh copy, run the nightly installer from https://nightly.multitheftauto.com and make sure to install all of the resources, open the game and click Map editor, press <kbd>F</kbd>, click Object (at the bottom-left), search for `321` or `Regular_Dildo`, click OK, place it, and click Current Elements (at the top). Optionally click Save As (at the top) to save the map and open the .map file in the \server\mods\deathmatch\resources\ folder in your MTA installation.

**Expected behavior**

The objects should be named like `object (gun_dildo1) (1)` and `object (gun_dildo2) (1)` and so on instead of `object (1)` and `object (2)` and so on.

**Version**

1.6.0-9.22953

q8X:
Fixes #601


SpeedyFolf:
I think you should've used the model names instead of custom names, just like every other object

Fernando-A-Rocha:
> I think you should've used the model names instead of custom names, just like every other object

FYI in the client objects.xml list for the Object Browser, multiple keywords are available for all weapon models

https://github.com/multitheftauto/mtasa-resources/blob/03c08613e2fac53328d996ff9c81f7feffd5a1c1/%5Beditor%5D/editor_gui/client/browser/objects.xml#L2457

q8X:
> I think you should've used the model names instead of custom names, just like every other object

i used engineGetModelNameFromID to get the names of the missing objects, feel free to make PR to change these if u want

PR #605, opened 2025-03-01
Update race_client.lua

Santiii733:
Fixed checkpoints problem from specting mode.

Fixed code by robson from MTA Race community

### DIFF
--- a/[gamemodes]/[race]/race/race_client.lua
+++ b/[gamemodes]/[race]/race/race_client.lua
@@ -1259,8 +1259,21 @@ function createCheckpoint(i)
 	if checkpoint.type == 'ring' and i < #g_Checkpoints then
 		setMarkerTarget(checkpoint.marker, unpack(g_Checkpoints[i+1].position))
 	end
-	checkpoint.blip = createBlip(pos[1], pos[2], pos[3], 0, isCurrent and 2 or 1, color[1], color[2], color[3])
+	checkpoint.blip = createBlip(pos[1], pos[2], pos[3], 0, isCurrent and 2 or 1, color[1], color[2], color[3], 255)
 	setBlipOrdering(checkpoint.blip, 1)
+	return checkpoint.marker
+end
+
+function createRadarBlips(i)
+	local checkpoint = g_Checkpoints[i]
+	if checkpoint.marker then
+		return
+	end
+	local pos = checkpoint.position
+	local color = checkpoint.color or { 0, 0, 80 }
+
+	checkpoint.blip2 = createBlip(pos[1], pos[2], pos[3], 0, isCurrent and 2 or 1, color[1], color[2], color[3], 100)
+	
 	return checkpoint.marker
 end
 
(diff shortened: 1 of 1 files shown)


[gamemodes]/[race]/race/race_client.lua
```diff
@@ -1259,11 +1259,24 @@ function createCheckpoint(i)
 	if checkpoint.type == 'ring' and i < #g_Checkpoints then
 		setMarkerTarget(checkpoint.marker, unpack(g_Checkpoints[i+1].position))
 	end
-	checkpoint.blip = createBlip(pos[1], pos[2], pos[3], 0, isCurrent and 2 or 1, color[1], color[2], color[3])
+	checkpoint.blip = createBlip(pos[1], pos[2], pos[3], 0, isCurrent and 2 or 1, color[1], color[2], color[3], 255)
 	setBlipOrdering(checkpoint.blip, 1)
 	return checkpoint.marker
 end
 
+function createRadarBlips(i)
```
Comment (Lpsd) [inline [gamemodes]/[race]/race/race_client.lua:1267]: Why is this function defined but not used anywhere?

[gamemodes]/[race]/race/race_client.lua
```diff
@@ -1287,10 +1300,22 @@ end
...
+
+		if (#g_Checkpoints - i) >= 3 then
+			--destroyElement(blip2.blip2)
+			if isElement(blip3.blip2) then
+				if blip3.blip2 then
+					destroyElement(blip3.blip2)
+				end
+			end
+		end
+
```
Comment (Lpsd) [inline [gamemodes]/[race]/race/race_client.lua:1318]: These blips are not guaranteed to exist with the current logic, yet you aren't checking if they do exist.

What happens if the last checkpoint in the table was destroyed, but more than 2 checkpoints exist?

Also the inconsistent usage of terms `blip` and `checkpoint` is confusing. You're getting a checkpoint from `g_Checkpoints`, not a blip - so these vars should be `local checkpoint2`, `local checkpoint3`, etc.

Lpsd (commented):
> Fixed checkpoints problem from specting mode.

What does this mean? Can you describe the problem you're trying to fix?

Aside from that, the suggested changes are a little odd; a function definition which is never used, and accessing table indexes without verifying they exist - please refer to the code review comments.

[gamemodes]/[race]/race/race_client.lua
Comment (Lpsd) [reply [gamemodes]/[race]/race/race_client.lua:1318]: `--destroyElement(blip2.blip2)` we should also not be submitting commented out code

Santiii733:
> > Fixed checkpoints problem from specting mode.
> 
> What does this mean? Can you describe the problem you're trying to fix?
> 
> Aside from that, the suggested changes are a little odd; a function definition which is never used, accessing table indexes without verifying they exist, and commented out code - please refer to the code review comments.


The issue is about the destroyElement error warning in blips on debugscript, which occurs when you are respawning back into the race or spectating a player.

[gamemodes]/[race]/race/race_client.lua
Comment (Santiii733) [reply [gamemodes]/[race]/race/race_client.lua:1267]: Nowadays, we use 3 blips in the race community, to accurately see the next 3cp from a distance.

[gamemodes]/[race]/race/race_client.lua
Comment (Santiii733) [reply [gamemodes]/[race]/race/race_client.lua:1318]: The codebase from vanilla mta itself throws a warning on -destroyelement, so you should check how to fix that problem. Cause problems on specting mode

PR #608, opened 2025-03-09
 Fix typo and Lint: Indentation, shadowing

PrimelPrime:
Fixed a typo regarding default scale value
Fixed Indentation and shadowing of scale which is now currentScale


Dutchman101:
Thanks

PR #612, opened 2025-03-18
internetradio: enforce configs to be loaded as first

### LINKED ISSUE #611: internetradio: Warning due to script load order
On a brand new MTA server installation, internetradio outputs the following warnings on start:

```
[00:23:41] WARNING: [gameplay]/internetradio/handle_radio/SHandleRadio.lua:219: Bad argument @ 'addCommandHandler' [Expected string at argument 1, got nil]
[00:23:41] WARNING: [gameplay]/internetradio/handle_radio/SHandleRadio.lua:275: Bad argument @ 'addCommandHandler' [Expected string at argument 1, got nil]
```

If `SHandleRadio.lua` is loaded before `SRadioConfig.lua` then this will happen.

In meta.xml, do the regex patterns guarantee an order on wildcard results? If not, we need to explicitly load `SRadioConfig.lua` and exclude it from the regex pattern for server includes.

```
[00:27:06] MTA:SA Server v1.6-release-23069
[mta@outpost multitheftauto_linux_x64]$ cat /etc/os-release
NAME="Rocky Linux"
VERSION="9.5 (Blue Onyx)"
```

ds1-e:
Fixes #611


[gameplay]/internetradio/meta.xml
```diff
@@ -2,6 +2,14 @@
 	<info author="MTA contributors (github.com/multitheftauto/mtasa-resources)" version="1.0" name="Speaker system" description="Internet radio player in /speaker" type="script"/>
 	<min_mta_version server="1.6.0-9.22430.0" client="1.3.4-0.00000"/>
 
+	<!-- those have to be loaded first -->
```
Comment (Lpsd) [inline [gameplay]/internetradio/meta.xml:5]:
```suggestion
	<!-- Load configs (must be loaded first) -->
```

[gameplay]/internetradio/meta.xml
```diff
@@ -2,6 +2,14 @@
...
+
```
Comment (Lpsd) [inline [gameplay]/internetradio/meta.xml:6]:
```suggestion
```

[gameplay]/internetradio/meta.xml
```diff
@@ -2,6 +2,14 @@
...
+	<script src="config/Sh*.lua" type="shared" cache="false"/>
+	<script src="config/C*.lua" type="client" cache="false"/>
+	<script src="config/S[!h]*.lua" type="server"/>
+
+	<!-- -->
```
Comment (Lpsd) [inline [gameplay]/internetradio/meta.xml:11]:
```suggestion
	<!-- Load logic scripts -->
```

[gameplay]/internetradio/meta.xml
```diff
@@ -2,6 +2,14 @@
...
+
```
Comment (Lpsd) [inline [gameplay]/internetradio/meta.xml:12]:
```suggestion
```

Lpsd (changes requested):
Looks fine apart from the comments.

I've tested this, and it does produce a warning about scripts already being included, due to using globstar. This isn't really an issue, but it it's annoying.

Unfortunately I don't think this can be resolved with our current folder structure. Can you move all folders except `config` under a new folder called `logic`? So the structure would be:

```
meta.xml
config/
    SRadioConfig.lua
    CRadioConfig.lua
    ShRadioConfig.lua
logic/
    gui/
        CRadioGUI.lua
    settings/
        CRadioSettings.lua
... etc
```

Then we can update the logic script includes, for example:
```xml
<script src="logic/**/Sh*.lua" type="shared" cache="false"/>
<script src="logic/**/C*.lua" type="client" cache="false"/>
<script src="logic/**/S[!h]*.lua" type="server"/>
```

This will avoid the duplicate script warning.

Lpsd (approved):
Thanks!

PR #630, opened 2025-04-24
Update delayindicator_client.lua

AlfredoAnonym:
Smaller delay indicator


PlatinMTA:
why are we using g_Me and g_Root? Why not root and localPlayer?

AlfredoAnonym:
> Thanks for your contributions, but it's not the best approach to submit a separate pull request for every single commit. Please take a moment to review the standard usage of Git, and think about combining similar changes into one pull request per topic.

Hey there!

Thank you for the suggestion. I did this the very first time, and I'm sorry if I spammed you LOL.

PR #646, opened 2025-07-15
Admin: add glitches and worldspecialproperties, fix lint error

PrimelPrime:
As title above this adds all worldpsecialproperties to the admin panel as well as glitches

https://wiki.multitheftauto.com/wiki/SetWorldSpecialPropertyEnabled
https://wiki.multitheftauto.com/wiki/SetGlitchEnabled

grafik

### DIFF
--- a/[admin]/admin/client/gui/admin_main.lua
+++ b/[admin]/admin/client/gui/admin_main.lua
@@ -18,6 +18,8 @@ aPlayers = {}
 aBans = {}
 aLastSync = 0
 aResources = {}
+glitchesData = {}
+worldPropsData = {}
 
 local serverPassword = 'None'
 local hasResourceSetting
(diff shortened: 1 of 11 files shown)


Dutchman101:
Please add a questionmark ('?') icon next to each listed glitch, that spawns a new GUI box with extended details about what the glitch exactly does and affects.

Also, please make this box appear whenever the Admin user is about to toggle a glitch, but with an additional "Are you sure to [Enable/Disable] this?" warning text, so that they are prompted to read the details first. Clicking Yes or No should close the box as well as perform or not perform the toggle.

Use a dark outline box with white text field that contains the details (Static text, can't be modified but can be selected and CTRL+C'd). The box must have a close button in the same format as other admin panel dialogs.
@PrimelPrime

PR #649, opened 2025-07-16
edf: Fix edf element falls or teleport when interacting with water/players

### LINKED ISSUE #639: editor: edf element falls/teleport when intreacting with water
edf

**Describe the bug**
the attached element is making the EDF elements fall and for some reason its a dynamic object which is causing this.
setting the position or freezing it doesn't help, but changing physical property solve this

https://github.com/multitheftauto/mtasa-resources/blob/7e86616099dc07e94bf05ad60010fc6edb1ab471/%5Beditor%5D/edf/edf.lua#L440-L462

i may make a PR to change the model to static object instead of current dynamic object

**To Reproduce**

1. add race gamemode to definitions
2. spawn race pickup
3. grab it and place it in the water
4. see element falling

another easy way to reproduce this is to make the player hit the race pickup using f6 mode

**Expected behavior**
elements shouldn't either fall or teleport when hitting it

**Version**
Multi Theft Auto v1.6-release-23196

**Additional context**

https://github.com/user-attachments/assets/3a288c68-541a-4860-b309-6ea141d77311

https://github.com/user-attachments/assets/93708edc-cb95-45c3-a9dc-bc50b41cab85

q8X:
Fixes #639, i changed the model from dynamic to static in order to fix this issue

old model
3003_w
new model
3045_w

### DIFF
--- a/[editor]/edf/edf.lua
+++ b/[editor]/edf/edf.lua
@@ -8,7 +8,7 @@ local thisResource = getThisResource()
 createResourceCallInterface("mapmanager")
 addEvent"onElementPropertyChanged"
 
-local DUMMY_ID = 3003
+local DUMMY_ID = 3045
 local DUMMY_DIMENSION = -99
 local DUMMY_INTERIOR = 14
 


q8X:
Found a new bug related to old dummy, new dummy fixes this bug

https://github.com/user-attachments/assets/0ed8620b-e5ef-4971-9922-81500f14772f

PR #655, opened 2025-07-30
race: add new settings to control vehicle physics behavior

xLive:
This PR adds new settings that can be configured globally by the server or per map to control vehicle physics behavior. By default, they do not change the current behavior unless set in race or map settings. These settings were added to fix issues caused by multitheftauto/mtasa-blue#4243, which broke many custom maps and made them unfinishable. They allow server owners and mappers to bring back the old physics on specific maps by disabling the fix and setting the FPS those maps were designed for, without affecting other maps. The old behavior can also be enabled globally.

**New settings:**
- **`vehicle_physics_mode`**  
Controls physics behavior. `fixed` for FPS-independent physics, or `legacy` to restore the old behavior. Default is `fixed`.
- **`vehicle_physics_fps`**  
Sets the FPS limit when `vehicle_physics_mode` is set to `legacy`. `-1` to keep the current limit, or set a specific FPS value. The original FPS limit is restored when the map or race resource stops. Default is `-1`.
- **`vehicle_physics_mode_map_can_override`** and **`vehicle_physics_fps_map_can_override`**  
Allow maps to override the global `vehicle_physics_mode` and `vehicle_physics_fps` values. Default is `true`.

### DIFF
--- a/[gamemodes]/[race]/race/edf/race.edf
+++ b/[gamemodes]/[race]/race/edf/race.edf
@@ -7,6 +7,8 @@
 	<setting name="respawntime" friendlyname="Respawn delay" type="natural" description="Time from death to respawn in seconds. Default is 5." default="5" required="true"/>
 	<setting name="autopimp" friendlyname="Auto pimp" type="selection:true,false" default="true" description="Whether to apply a selection of visual upgrades to vehicles." required="false"/>
 	<setting name="firewater" friendlyname="Fire water" type="selection:true,false" default="false" description="Whether to explode the player when entering the water." required="false"/>
+	<setting name="vehicle_physics_mode" friendlyname="Vehicle physics mode" type="selection:fixed,legacy" description="Sets the vehicle physics mode. Use fixed (recommended) for FPS-independent physics, or legacy to use FPS-dependent physics." required="false"/>
+	<setting name="vehicle_physics_fps" friendlyname="Vehicle physics FPS" type="string" description="FPS limit to use when vehicle physics mode is set to legacy. This setting changes the limit: -1 = leave unchanged, 0 = unlimited, 25+ = specific limit." required="false"/>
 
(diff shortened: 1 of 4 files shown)


Dutchman101:
This is my proposal for a perfect design:
- If a racing map has no physics (legacy or fixed) flag in meta.xml / Map settings, default to legacy. So that old maps need no intervention to provide the same experience as which they were designed for.
- When a racing map starts that has a specific physics flag in meta.xml / Map settings, apply it accordingly.
- Make sure that if "race" gamemode stops, it restores the glitch state to what it was when "race" loaded, or what it was before the first map that "race" decided a physics setting for (based on the map properties).

So then, racing servers only need to update mtasa-resources when MTA 1.7 releases in the coming few weeks, something that has been stressed with every new release anyways.

xLive:
> This is my proposal for a perfect design:
> 
> * If a racing map has no physics (legacy or fixed) flag in meta.xml / Map settings, default to legacy. So that old maps need no intervention to provide the same experience as which they were designed for.

When a racing map hasn’t set the physics mode, it will use whatever the server has set in the race resource settings. By default, it was `fixed`, but I’ve now changed it to `legacy`.

> * When a racing map starts that has a specific physics flag in meta.xml / Map settings, apply it accordingly.
> * Make sure that if "race" gamemode stops, it restores the glitch state to what it was when "race" loaded, or what it was before the first map that "race" decided a physics setting for (based on the map properties).

The PR already does both of these. It applies the new settings based on what each map has in its meta settings. If the map hasn’t set anything, it uses the values from the race resource settings. It also restores the glitch state when the race resource stops.

_Side note_: I suggest we use `fixed` physics mode for the default race maps to benefit from the physics fix. From quick testing on several maps, it’s very unlikely they will be _broken_ by `fixed` mode. If that’s okay, I will change them in a future PR.

Issue #559, opened 2024-10-13
Missing xmlns namespace declaration when saving map


CrosRoad95:
**Which resource(s) have this problem?**

Map editor

**Describe the bug**

When saving map, map editor don't add xmlns namespace declaration.

**To Reproduce**
Create any map, save it. you will see:
```xml
<map edf:definitions="editor_main"></map>
```
as your map.

**Expected behavior**
Root node should contain namespace definition:
```xml
<map xmlns:edf="https://wiki.multitheftauto.com/wiki/Resource:Editor/EDF" edf:definitions="editor_main">
```
**Screenshots**
/

**Version**
all mta versions

**Additional context**

PR #658, opened 2025-08-14
editor_main: Add xmlns:edf namespace when saving maps

omar-o22:
Fix: #559  
This PR adds the `xmlns:edf` namespace attribute to the map node when saving maps.


Dutchman101:
Thanks!
@omar-o22

PR #675, opened 2025-09-07
Refactor reload script to server-side only

shadylua:
This PR refactors the reload script to run server-side only, removing the client-side implementation.


jlillis (changes requested):
I think this would introduce new problems than improvements:

1 - By moving all the logic server side, we lose the ability to block reload during certain tasks (`getPedSimplestTask` is client-only).
2 - By moving the `bindKey` call server side, players lose the ability to change the keybind via the MTA settings menu.

I think the current approach with both client and server logic works best.

PR #677, opened 2025-09-07
editor_main: Fix permissions after #4388

q8X:
This PR fixes an issue introduced by https://github.com/multitheftauto/mtasa-blue/pull/4388


PerikiyoXD:
This seems to bypass ACL right enforcement.

I'd advocate for proper ACL right handling per group by default on `editor_acl.xml`

PR #684, opened 2025-09-18
admin2: Add GUI panel for mute

### LINKED ISSUE #108: admin2: permanent mute from serial
I would like to mute abusive players. Permanently.

An option to mute players permanently from the menu. Based on serial or IP.

**Details**
We would need a way to store serial and IP mutes as well as a way to track the associated player (just list the name that was banned?)

**Additional context**
Request received from [forum](https://forum.mtasa.com/topic/109848-permanent-mute-serial-mute/) seems reasonable.

Please add "permanently from IP" and "permanently from serial".
image

omar-o22:
Summary
-
- This PR adds a proper mute system with a new GUI and fixes #108.  
- Previously, mutes were just a toggle and expired on reconnect.  
- Now, admins can set a duration, and the mute duration only counts while the player is connected (it pauses if the player disconnects).

Preview:
-
image

Edited
-
I did a mention for issue 106 by mistake


ArranTuna (changes requested):
If you close admin panel while the mute GUI is open, it doesn't close, and then after that I clicked close (with console) and GUI won't close.

Fernando-A-Rocha:
I agree mute duration should be minutes by default, but also allow other units of measurement for custom needs (seconds, hours..)

ArranTuna:
You're right, what you have to do is, press mute twice, then close the panel and you're stuck with a second panel. So basically when you click mute, don't make another GUI if one already exists.

jlillis (dismissed):
I don't think this quite resolves #108, as it mutes are not able to persist if the server or resource restarts. There will also need to be a new tab to view, modify, and delete existing mutes.

ArranTuna:
> > You're right, what you have to do is, press mute twice, then close the panel and you're stuck with a second panel. So basically when you click mute, don't make another GUI if one already exists.
> 
> I don't know where to save the data because admin2 saves some data in conf/***.xml and some in admin.db too So should I save data in admin.db or create a new xml file?

Use this existing connection to admin.db:

connection = dbConnect("sqlite", "admin.db"),

And use the dbExec and dbQuery functions and use dbQuery call back function feature.

[admin]/admin2/server/admin_mute.lua
```diff
@@ -0,0 +1,273 @@
...
+	local serial = getPlayerSerial( player )
+    return aMutedList[serial]
+end
+
+function aAddOrUpdateMute(serial, player, admin, reason, time, isRestarted)
+    local result
+    local query_text = "SELECT serial FROM mutes WHERE serial = ?"
+    if ( isRestarted ) then
+        local query = dbQuery( db.connection, query_text, serial )
+        result = dbPoll( query, -1 )
```
Comment (ArranTuna) [inline [admin]/admin2/server/admin_mute.lua:194]: You should be using the call back feature.

ArranTuna:
I added a mute on myself and don't see it in mutes tab and refresh button is disabled?

Lpsd (approved):
Looks good to me.

As mentioned in Discord, may be nice to have an option/checkbox for muting player voice too, if that's possible.

ArranTuna (approved):
I will just make the change myself about using the db call back

PR #686, opened 2025-09-19
editor_gui: Fix broken case-insensitive search

q8X:
previously the old method was making the search term lowercase leaving the map name in the original case


Nico8340 (approved):
LGTM

PR #687, opened 2025-09-19
editor_gui: Fix overlapping checkboxes

q8X:
Fix an issue where clicking on "enable sounds" or "enable bounding box" causes overlapping

Before:

After:


Nico8340 (approved):
LGTM

PR #688, opened 2025-09-21
admin: Fix admin slap not respawning players on fatal damage

MohabCodeX:
Admin panel slap with fatal damage kills players but they don't respawn automatically.

Solution: Changed from forced kill to natural health reduction, allowing normal respawn mechanics to work.


ArranTuna:
There's nothing wrong with admin panel, I have checked 1.6 and 100 HP slap does respawn you, there is a bug in MTA 1.7 not admin panel.

PR #689, opened 2025-09-25
Map Editor: highlight cursor

x-Crown:
mta-screen_2025-09-26_02-34-39
This PR introduces an additional hover highlight effect in the MTA:SA Editor. When the player hovers over an object in the map editor:

The object becomes semi-transparent.

A 3D animated bounding box is drawn around it with rainbow cycling colors (HSV → RGB).

The effect resets automatically when the cursor leaves the object


q8X:
i was going to implement this when this PR came out, but it seems like one of the recent changes to editor broke settings (kinda)

https://github.com/multitheftauto/mtasa-resources/blob/604c05e9f1d3e72fd1fad5757b008eefa6002511/%5Beditor%5D/editor_gui/client/options_backend.lua#L133-L137

this issue could delete new settings if the mapper switches to outdated editor (which is likely going to happen when mappers switch between servers)

PR #690, opened 2025-09-28
superman: fix bug

omar-o22:
- Steps to reproduce
1. Enable superman using /superman or double shift
2. Click on F1 > kill or set your HP to 0 while moving
3. Your ped can no longer move.

In this PR this bug fixed.


Nico8340 (approved):
LGTM

PR #692, opened 2025-10-04
[editor] Improved object preview

chris1384:
fixes:
- fixed object not appearing when selecting certain model ids
- object rotation in viewport is centered
- removed zoom out limit
- added doublesided property to previewed element


Fernando-A-Rocha:
Nice but some objects aren't supposed to be double sided?

chris1384:
you're right, I thought about the outcomes of it, but I considered that there are more objects that need easier viewing (mostly landbits and vegetation) and there are only a few models that might look out of place when doublesided is applied to them

we'll have to see if it's required to either keep it or remove it

PR #696, opened 2025-11-07
Add "Lua time recordings" to web performancebrowser

Xenius97:
This Pull Request adds tracking for high CPU usage resources to the web-based performance browser. Similar to the IPB, everything can now be monitored even while you're offline.
If you don’t want to use it, you can disable it - based on my experience, it doesn’t negatively impact performance.

image

### DIFF
--- a/[admin]/admin/client/gui/admin_main.lua
+++ b/[admin]/admin/client/gui/admin_main.lua
@@ -1408,7 +1408,7 @@ end
 
 function setAnonAdmin( bOn )
 	guiCheckBoxSetSelected ( aTab1.AnonAdmin, bOn )
-	setElementData( localPlayer, "AnonAdmin", bOn )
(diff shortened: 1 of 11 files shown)


[web]/performancebrowser/viewer.lua
```diff
@@ -155,6 +155,9 @@ function Viewer:getCategoriesRaw ()
 	local categories = {}
 	for _,row in ipairs(rowList) do
 		table.insert( categories, row[1] )
+		if ( row[1] == "Lua timing" ) then -- Add our custom Lua time recordings category
```
Comment (jjustns) [inline [web]/performancebrowser/viewer.lua]:
```lua
if ( row[1] == "Lua timing" and target.bIsServer ) then -- Add our custom Lua time recordings category
```

Comment (Xenius97) [reply [web]/performancebrowser/viewer.lua]: Done, thank you!

PR #699, opened 2025-11-29
admin/admin2: Refactor use tables instead of element data

omar-o22:


omar-o22:
> But aren't we meant to improve "admin2" and not "admin" resource?

It's not only an improvement it also fixes security backdoors, cheaters can currently toggle anonymous mode for admins
Btw I'll convert this PR for draft because I'm currently fix this issue in admin2 (I forget)

PR #704, opened 2025-12-24
Map Editor: A feature for simple moving objects

IIYAMA12:
Allows you to make some simple object movements in your map editor.

Changes:
- More object properties
- With the prop persistDefault="false", the data field will not persist when a default value is provided. Else this update would make all maps a bit bigger. I know persisting default values is generally better, but for this user case I find it unnecessary
- Editor preview of how the objects are moving from start to end
- Editor extension (file) modified to move the objects

Unknowns:
- [x] What does the edf.edf file do? I am not sure if I have to keep that file in sync with the editor_main.edf file.

mta-screen_2025-12-24_16-32-33
mta-screen_2025-12-24_17-29-39

### DIFF
--- a/[editor]/editor_main/client/gridlines.lua
+++ b/[editor]/editor_main/client/gridlines.lua
@@ -155,7 +155,7 @@ function drawObjectMoveLines()
 
 		local fullCycle = 2 * time + 2 * delay
 		local cyclePos = timeNow % fullCycle
-		local progress = 0
+		local progress
 		if cyclePos < delay then
 			-- Delay at 0
 			progress = 0


Lpsd (commented):
This is a solid addition to the map editor, thank you!

I've tested the changes, they're working well. Although I do wonder whether it's worth adding an extra setting for "loop" to decide whether to repeat the movement at all? Currently an object will just keep looping.

Regarding your question on edf.edf, it's not necessary to update it here. The .edf file provided by the "edf" resource determines the default definitions for other resources (a fallback, when they don't provide their own .edf). Since your feature is entirely within the map editor, and bundles functionality into the "mapEditorScriptingExtension_s.lua" script (which is ran by gamemodes/resources such as race on map load), other resources don't need to know about these properties - they aren't "standard" properties.

Essentially, you only need to update the edf.edf for new "standard" entity properties, that all resources should be aware of. No need to update it here!

PR #729, opened 2026-06-26
Optimise PNG images

SpeedyFolf:
Optimised PNG images because apparently people forgot to


qaisjp:
I thought we already had a pull request like this several years ago.

SpeedyFolf:
Yea, but stuff was updated since then. This doesn't modify every image.

Lpsd:
#240 is the relevant PR

Looks like most the files modified in this PR were either not included in #240, didn't exist at that time, or have been updated with less efficient versions since then.

PR #734, opened 2026-08-04
Add licenseplate resource

DateFTP:
**Type:** New resource

**Summary**
Adds a new `[gameplay]/licenseplate` resource that lets players set a custom license plate on the vehicle they are currently driving through a small in-game GUI window.

**What the resource does**
- **Client (`client.lua`)**
  - Creates a small GUI window ("License plate") with a text field, a "Set the number" button and a "Cancel" button.
  - Opens the window with the commands `/setplate` and its alias `/license`.
  - Validates the input on the client side:
    - rejects an empty plate with a chat message;
    - rejects plates longer than 8 characters (matching the vanilla plate limit);
    - an optional letter/number/space-only check is provided but commented out.
  - Checks that the player is actually inside a vehicle before continuing.
  - Triggers the server-side event `LicensePlate:set` with the vehicle and the plate, then hides the window.
- **Server (`server.lua`)**
  - Listens for the `LicensePlate:set` event.
  - Validates the player, the vehicle and that the player is still inside that vehicle.
  - Enforces the 8-character limit again server-side (trusted check).
  - Applies the plate with `setVehiclePlateText` and confirms with a chat message.

**Files**
- `client.lua` - GUI creation, input validation, commands and event triggering.
- `server.lua` - validation and applying the plate via `setVehiclePlateText`.
- `meta.xml` - resource definition, registers the client and server scripts.

**Tested**
- Opening/closing the window via both commands.
- Empty input, oversized input and a valid plate.
- Setting a plate while not in a vehicle.

### DIFF
--- a/[gameplay]/licenseplate/client.lua
+++ b/[gameplay]/licenseplate/client.lua
@@ -1,0 +1,89 @@
+local GUI = {
+    button = {},
+    window = {},
+    label = {},
+    edit = {}
+}
+
+local isWindowOpen = false
+
+function createLicenseWindow()
+    if GUI.window[1] then return end 
+
+    GUI.window[1] = guiCreateWindow(671, 319, 307, 152, "License plate", false)
+    guiWindowSetSizable(GUI.window[1], false)
+
+    GUI.edit[1] = guiCreateEdit(13, 46, 284, 44, "", false, GUI.window[1])
+    GUI.label[1] = guiCreateLabel(16, 23, 281, 19, "Specify the number you want to install.", false, GUI.window[1])
+    guiSetFont(GUI.label[1], "default-bold-small")
+
+    GUI.button[1] = guiCreateButton(9, 104, 129, 38, "Set the number", false, GUI.window[1])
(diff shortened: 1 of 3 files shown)


DateFTP:
image

DateFTP:
image
image

T-MaxWiese-T:
In my opinion, a standard resource should be as customizable as possible, so I suggest there should be an `export` function that allows the server to restrict, for each vehicle, which players are allowed to change the license plate.

DateFTP:
I will consider the possibility of adding exports of functions and configs — August 5 is included until the end of the day. Thank you for the feedback regarding exports and settings.

qaisjp:
thank you for your contribution. I think it would make sense to submit this to the MTA Community website instead, as we have a very high bar for new default resources, and this resource is limited in the value it can provide to server owners.

This feature would probably be fine to add to the freeroam resource, but we will not be accepting this as a brand new resource.

(If another maintainer disagrees, I'm happy to chat about it -- we can always reopen the pull request later.)

PR #736, opened 2026-08-08
Parachute2 

RushCuttle:
What was done so far:

- forwards/backwards/stall in skydive mode are close to SA
- When parachute opens, it mimicks SA 
- Death when hitting buildings
- Death when hitting objects 
- Animations work 

Issues: 

- If you spam W/S you'll be able to do the walk animation during skydive mode
- Fowards to backwards and vice versa is a little bit awkward (feels a little bit robotic not perfect) 

TODO list: 

- Strafe while in Skydive mode is not polished 
- Strafe while in parachute mode is not polished (needs to add the pitch for when strafing)
- Add the exit parachute logic
- Sync (was commented out)
- clean the code and rework a little bit of the logic

How it's today: 
- https://youtu.be/Tl1RJzPaBNA


ArranTuna:
Since this is not completed, I've marked it as draft, when complete, mark it as not a draft.

PR #738, opened 2026-08-11
editor: Fix editor breaking on old race maps with legacy rotation

q8X:
old race maps store rotation as a single `rotation="0"` attribute instead of separate rotX/Y/Z, this fix should help old maps open and save without errors

before:

https://github.com/user-attachments/assets/b5811222-c47e-4f96-9366-a5a5ab9750c6

after:

https://github.com/user-attachments/assets/59b022df-da5a-4fc8-8dee-3fb1ab8ad9d6

### DIFF
--- a/[editor]/edf/edf.lua
+++ b/[editor]/edf/edf.lua
@@ -858,7 +858,7 @@ function edfGetElementPosition(element)
 	end
 end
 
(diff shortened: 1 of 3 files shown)


vectorbasis (approved):
LGTM :shipit:

PR #745, opened 2026-09-02
Fix event handler target root, nil element destruction, and type conversions in interiors

iManGaaX:
- Updated `doWarpPlayerToInterior` event handler in `interiors_client.lua` to target `root` instead of `localPlayer` so server triggers execute properly.
- Added `isElement` checks before destroying markers and colshapes during resource cleanup to avoid runtime errors on one-way or missing markers.
- Safe-cast coordinates, dimensions, and rotation parameters with `tonumber` to prevent arithmetic type mismatch errors when calculating player rotation.
- Implemented `isElement(player)` validation inside asynchronous timer callbacks (`setPlayerInsideInterior`) to prevent server warnings if a player quits mid-warp.

### DIFF
--- a/[gameplay]/interiors/interiorexp.lua
+++ b/[gameplay]/interiors/interiorexp.lua
@@ -69,11 +69,11 @@ local function interiorCreateMarkers ( resource )
 				setElementParent ( marker, entryInterior )
 				interiorMarkers[entryInterior] = marker
 				--
-				local dimension = tonumber(getElementData ( entryInterior, "dimension" )) or 0
-				local interior = tonumber(getElementData ( entryInterior, "interior" )) or 0
+				local entDimension = tonumber(getElementData ( entryInterior, "dimension" )) or 0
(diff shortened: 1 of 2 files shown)


ArranTuna:
There's some lint errors to be fixed.

iManGaaX:
> There's some lint errors to be fixed.

It should now be done.

ArranTuna:
You didn't fix anything, you just completely broke the script instead, because you didn't even bother to try starting the script, let alone actually test it in-game:

[2026-09-07 11:35:24] Starting interiors
[2026-09-07 11:35:24] ERROR: [gameplay]\interiors\interiorexp.lua:16: attempt to call global 'interiorLoadElements' (a nil value)
[2026-09-07 11:35:24] start: Resource 'interiors' started

Let me guess, an AI decided to make that function local, but it didn't even verify if that function was being called above or in another resource.

PR #750, opened 2026-09-20
Restore meta.xml content to original state

NightWatch404:
- Ping Kicker: 0-10000
- FPS Kicker: 0-100
- Idle Kicker: 0-6000

No default values or runtime behavior were changed


ArranTuna:
Pointless.

PR #758, opened 2026-09-20
Restrict remote admin chat messages

NightWatch404:
The remotely triggerable `aAdminChat` event currently accepts messages without verifying that the caller has admin chat permission, allowing unauthorized clients to inject messages into admin chat Validate `general.tab_adminchat` and use client as the sender


ArranTuna:
Although this was needed as hasObjectPermissionTo check was needed, the "client or source" was needed to allow the admin chat command to work, so I've fixed it.

